Issue #20, opened 2022-08-22
Hackweek TODOs


lforst:
## Stage 0 - Prelude

- [x] Uplugin explanation session
- [x] Repo explanation session

## Stage 1 - MVP

- [x] Finish up / rethink release injection mechanism (https://github.com/getsentry/hackweek-sentry-unplugin/pull/24)
- [x] Write function to get a proper `release` value (https://github.com/getsentry/hackweek-sentry-unplugin/pull/21, https://github.com/getsentry/hackweek-sentry-unplugin/pull/22)
- [x] Source maps upload 1.0  (https://github.com/getsentry/hackweek-sentry-unplugin/pull/23)
  - [x] Do what Sentry CLI does while uploading source maps (in MVP form).  (https://github.com/getsentry/hackweek-sentry-unplugin/pull/26)
- [x] Source maps Upload 2.0
  - [x] Replace Sentry-CLI calls step by step:
    - [x]  (required) create release (https://github.com/getsentry/hackweek-sentry-unplugin/pull/29)
    - [x] (required) cleanArtifcats (https://github.com/getsentry/hackweek-sentry-unplugin/pull/30)
    - [x] (required) upload sourcemaps (https://github.com/getsentry/hackweek-sentry-unplugin/pull/33)
    - [x] (required) finalize release (https://github.com/getsentry/hackweek-sentry-unplugin/pull/32)
- [x] Handle `entries` option (https://github.com/getsentry/hackweek-sentry-unplugin/pull/38, ref: https://github.com/getsentry/hackweek-sentry-unplugin/issues/28)
- [x] Remove facade (#46)

## Stage 2 - Ship

- [x] Get license figured out (https://github.com/getsentry/hackweek-sentry-unplugin/issues/5)
- [x] Check package jsons (implicitly done in https://github.com/getsentry/hackweek-sentry-unplugin/pull/42)
- [x] Write READMEs (#39)
- [ ] Add examples to readmes
- [ ] Add CHANGELOG
- [ ] Remove any console logs that are not explicit debug logging
- [ ] Publish package

## Stage 3 - Feature parity with webpack plugin
- [ ] Extend basic functionality with all options from webpack plugin
- [ ] add missing commands for release creation pipeline:
    - [ ] set commits  
    - [ ] add deploy
- [ ] To be defined

## Misc Todos:

- [x] Write integration tests (https://github.com/getsentry/hackweek-sentry-unplugin/pull/25)
[...]

PR #21, opened 2022-08-22
chore: add release discovery

vladanpaunovic:
relates: #20  

This behaviour tries to replicate the behaviour of release option of sentry-webpack-plugin.

Release name can be specified via process.env.SENTRY_RELEASE. Otherwise, it will automatically detects values for Heroku, AWS CodeBuild, CircleCI and Vercel. Otherwise uses HEAD's commit SHA. (For HEAD option, requires access to git CLI and for the root directory to be a valid repository).

Logic taken from [sentry-cli here](https://github.com/getsentry/sentry-cli/blob/master/src/utils/releases.rs#L87-L143)

### DIFF
--- a/packages/unplugin/src/getReleaseName.ts
+++ b/packages/unplugin/src/getReleaseName.ts
@@ -1,0 +1,38 @@
+import * as child_process from "child_process";
+import * as fs from "fs";
+import * as path from "path";
+
+function isGit(dir: string) {
+  return fs.existsSync(path.join(dir, ".git"));
+}
+
+function getBranchHead() {
+  return child_process.execSync("git rev-parse HEAD").toString().trim();
+}
+
+export function getReleaseName(releaseName?: string): string {
+  if (releaseName) {
+    return releaseName;
+  }
+
+  const ENV_VARS = [
+    "SENTRY_RELEASE",
+    "SOURCE_VERSION", // Heroku #1 https://devcenter.heroku.com/changelog-items/630
+    "HEROKU_SLUG_COMMIT", // Heroku #2: https://docs.sentry.io/product/integrations/deployment/heroku/#configure-releases
+    "CODEBUILD_RESOLVED_SOURCE_VERSION", // AWS CodeBuild: https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+    "CIRCLE_SHA1", // CircleCI: https://circleci.com/docs/2.0/env-vars/
+    "VERCEL_GIT_COMMIT_SHA", // Vercel docs: https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables
+  ];
+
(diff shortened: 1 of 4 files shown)


packages/unplugin/src/getReleaseName.ts
```diff
@@ -0,0 +1,38 @@
...
+    return releaseName;
+  }
+
+  const ENV_VARS = [
+    "SENTRY_RELEASE",
+    "SOURCE_VERSION", // Heroku #1 https://devcenter.heroku.com/changelog-items/630
+    "HEROKU_SLUG_COMMIT", // Heroku #2: https://docs.sentry.io/product/integrations/deployment/heroku/#configure-releases
+    "CODEBUILD_RESOLVED_SOURCE_VERSION", // AWS CodeBuild: https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+    "CIRCLE_SHA1", // CircleCI: https://circleci.com/docs/2.0/env-vars/
+    "VERCEL_GIT_COMMIT_SHA", // Vercel docs: https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables
```
Lms24, line 24: Does the order matter here? If yes, is the order correct?
(Looking at it again, I think it doesn't but I'm not sure about Heroku...)

packages/unplugin/test/getReleaseName.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  it("adheres to HEAD when git is present", () => {
+    (mockedFs.existsSync as jest.Mock).mockReturnValueOnce(true);
+    const sha = "c3f235fc86f1c4007e3a218ec82d666586e73cbf";
+    (mockedChildProcess.execSync as jest.Mock).mockReturnValue(sha);
+
+    expect(getReleaseName()).toBe(sha);
+  });
+
+  it("throws an error if git is not present", () => {
```
Lms24, line 34: nit: I think this better describes the behaviour because this is the "last resort" if we have nothing to determine the release.
```suggestion
  it("throws an error if no release information could be found", () => {
```

Lms24 (approved):
Nice! 🚀

packages/unplugin/test/getReleaseName.test.ts
Lms24: And since it's hackweek: Feel free to ignore ;)

packages/unplugin/src/getReleaseName.ts
vladanpaunovic: Great question! It does. This is why `SENTRY_RELEASE` is on the top of the list, everything else is carrying equal weight. I was thinking to extract `SENTRY_RELEASE` from the rest of the pack but found it to be overkill

Lms24: Thanks for explaining. Makes sense! Then, let's make sure to add a comment that `SENTRY_RELEASE` should be the first item in the list.

PR #22, opened 2022-08-22
chore: extract SENTRY_RELEASE from rest of the group

vladanpaunovic:


Lms24 (approved):

Lms24:
I think that's better in terms of code readability. Good change!

PR #23, opened 2022-08-22
feat(sourcemaps): add SentryFacade and Sentry CLI Boilerplate

Lms24:
This PR marks the first step towards getting source maps and release creation/finalization working. It adds Sentry-CLI as a dependency, as well as a facade object for the necessary steps (see explanation below).

While we want to get rid of depending on Sentry CLI eventually, I decided to use it for now to get our MVP out faster. To replace Sentry CLI functionality easily with our "native" functionality later, I created a facade to abstract the implementation of the necessary steps from what we call in the "front end" of our plugin. We can replace each CLI command step by step with our own implementation later on.

Furthermore, I added a few fields to `Options`+ a bunch of commented out ones from the webpack plugin and I moved the type definition to a new `types.ts` file 

Next up: Implement CLI calls to:
* create a release
* upload sourcemaps
* finalize release

There are more CLI calls in the webpack plugin (clear artifacts, set commits, add deploy) but we don't necessarily have to add them to the MVP.


packages/playground/vite.config.js
```diff
@@ -14,5 +14,14 @@ export default defineConfig({
     },
     sourcemap: true,
   },
-  plugins: [sentryVitePlugin()],
+  plugins: [
+    sentryVitePlugin({
+      org: "sentry-sdks",
```
vladanpaunovic, line 19: nit: can we share common settings with other configs? it would be nicer if possible

packages/unplugin/src/index.ts
```diff
@@ -17,15 +19,12 @@ function generateGlobalInjectorCode({ release }: { release: string }) {
...
-}
-
 const unplugin = createUnplugin<Options>((options) => {
-  function debugLog(message: string) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function debugLog(...args: any) {
     if (options?.debugLogging) {
       // eslint-disable-next-line no-console
-      console.log(`[Sentry-plugin] ${message}`);
+      console.log("[Sentry-plugin]}", args);
```
vladanpaunovic, line 27: nit: there is an additional `}` here

vladanpaunovic (approved):
+1 for the idea of getting rid of sentry-cli later on

While it is hard to make a call on this PR as it seems like a WIP, I like the direction it is going so ✅  from me

packages/unplugin/src/index.ts
Lms24: Done [c5c5fec](https://github.com/getsentry/hackweek-sentry-unplugin/pull/23/commits/c5c5fec4619b1966f9c344392a0123b85da8b7ff)

packages/playground/vite.config.js
Lms24: Sure, tbh I'm not really happy with adding actual projects here. I just used this values to test out basic functionality (such as calling the CLI via the plugin). I can extract this to some sort of placeholder config object but I'll try to come up with something different

vladanpaunovic (approved):

PR #26, opened 2022-08-23
feat(sourcemaps): Add basic source maps upload functionality

Lms24:
* make Sentry CLI calls
* adds default options
* add small node sample app to test sourcemaps upload

### DIFF
--- a/packages/playground/config.json
+++ b/packages/playground/config.json
@@ -1,5 +1,4 @@
 {
-  "configFile": "./sentry.properties",
   "release": "0.0.1",
   "org": "some-org",
   "project": "some-proj",
--- a/packages/playground/config.json
+++ b/packages/playground/config.json
@@ -1,4 +1,5 @@
 {
+  "configFile": "./sentry.properties",
   "release": "0.0.1",
   "org": "some-org",
   "project": "some-proj",
--- a/packages/playground/src/smallNodeApp.js
+++ b/packages/playground/src/smallNodeApp.js
@@ -1,0 +1,31 @@
+const Sentry = require("@sentry/node");
+const { RewriteFrames } = require("@sentry/integrations");
+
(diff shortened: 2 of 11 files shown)


packages/playground/src/smallNodeApp.js
```diff
@@ -0,0 +1,31 @@
+const Sentry = require("@sentry/node");
+const { RewriteFrames } = require("@sentry/integrations");
+
+// import * as Sentry from '@sentry/node';
+// import { RewriteFrames } from "@sentry/integrations";
```
lforst, line 5: Is this still needed?

packages/playground/config.json
```diff
@@ -1,4 +1,5 @@
 {
+  "configFile": "./sentry.properties",
```
lforst, line 2: Is this still needed? As far as I can see we don't have a properties file

packages/playground/vite.config.smallNodeApp.js
```diff
@@ -0,0 +1,31 @@
+// @ts-check
+import { sentryVitePlugin } from "@sentry/unplugin";
+import { defineConfig } from "vite";
+import * as path from "path";
+import placeHolderOptions from "./config.json";
```
lforst, line 5: do we want to use the placeholderoptions in this file?

lforst (approved):

packages/playground/src/smallNodeApp.js
Lms24: nope :D

packages/playground/config.json
Lms24: Nope :D

packages/playground/vite.config.smallNodeApp.js
Lms24: Nope :D

PR #35, opened 2022-08-24
feat: Add `errorHandler` option

vladanpaunovic:
This PR adds a minimal implementation for `errorHandler`, exposing errors from release publishing to users.

Unlike the signature in `sentry-webpack-plugin`, this one is reduced to essentials as I to my understanding, the rest is webpack specific and a workaround on not throwing the error

The difference:
```diff
- errorHandler?: (err: Error, invokeErr: () => void, compilation: unknown) => void;
+ errorHandler?: (err: Error) => void;
```


Lms24 (dismissed):

vladanpaunovic:
Thanks @Lms24! You can read the rationale from the previous implementation [here](https://github.com/getsentry/sentry-webpack-plugin/blob/137503f3ac6fe423b16c5c50379859c86e689017/index.d.ts#L95-L114)

Lms24 (commented):
Okay, now a better review ;) 

So, this LGTM concept-wise but we need the invoke-error functionality:

> Function to call a when CLI error occurs. Webpack compilation failure can be triggered by calling invokeErr callback. Can emit a warning rather than an error (allowing compilation to continue) by setting this to (err, invokeErr, compilation) => { compilation.warnings.push('Sentry CLI Plugin: ' + err.message) }. Defaults to (err, invokeErr) => { invokeErr() }.

So the idea is: when I as a plugin user implement this `errorHandler` hook, I can decide if I want the bundling process to fail or just continue. In case I want it to fail, I call `invokeError` (else I don't).

So, the place where we call `options.errorHandler()` is correct. I suggest, we create a function `invokeError(e: Error)` in our code which will `throw e`. We pass this function as the second argument when calling `errorHandler`.

About `compilation` (the third argument): I'm not entirely sure what the point of this thing is. We likely need to drop it if unplugin doesn't provide something blunder-agnostic. Thoughts @lforst ? 
Regardless, I think we can revisit `compilation` later.

lforst:
Looked at this a bit. Apperently it's not really possible for us to do error handling like we do in the webpack plugin, since the unplugin doesn't provide us with a way of pushing to Webpack's `compilation.errors` and other bundlers don't really have an equivalent.

I'd say for now this is fine. But let's wait with merging this until we have Sentry merged since there are things to consider - we don't want errors thrown by user code to be sent to Sentry.

lforst (approved):

PR #36, opened 2022-08-25
feat: Add Sentry

vladanpaunovic:
This PR adds Sentry integration to sentry-unplugin.

This will allow us to proactively monitor and get alerted about our customers errors.

### Error page:
Screenshot 2022-08-25 at 02 31 49

### Performance:
Screenshot 2022-08-25 at 02 20 54

_Side note so I don't forget:
I would have expected to get distributed tracing when making requests to sentry API as we are sending `sentry-trace`. Let's discuss why this could happen. Different org maybe?_

### DIFF
--- a/packages/unplugin/src/index.ts
+++ b/packages/unplugin/src/index.ts
@@ -97,7 +97,12 @@ const unplugin = createUnplugin<Options>((originalOptions, unpluginMetaContext) 
     console.log("[Sentry-Plugin]", "To disable telemetry, set `options.telemetry` to `false`.");
   }
 
-  sentryHub?.setTags({ organization: options.org, project: options.project });
+  sentryHub?.setTags({
+    organization: options.org,
+    project: options.project,
+    bundler: unpluginMetaContext.framework,
+  });
+
   sentryHub?.setUser({ id: options.org });
 
   function debugLog(...args: unknown[]) {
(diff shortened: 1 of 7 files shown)


packages/unplugin/src/index.ts
```diff
@@ -3,6 +3,14 @@ import MagicString from "magic-string";
 import { getReleaseName } from "./getReleaseName";
 import { Options } from "./types";
 import { makeSentryFacade } from "./sentry/facade";
+import * as Sentry from "@sentry/node";
+import "@sentry/tracing";
+
+Sentry.init({
+  dsn: "https://1298ac2f87bc4aae99d9a9a94ce65b42@o447951.ingest.sentry.io/6681372",
```
vladanpaunovic, line 10: this dns is intentionally not an env var. I decided to keep it here as this is a public info and we should be transparent about this.

lforst: Agree!

lforst:
This PR is veeery cool. I love that we have tracing for the hooks 👌 

I have to look at this in more detail tho: My concern right now is that if we use the top-level `init` instead of having a custom client, we'll pollute other sentry instances if they're running.

lforst:
Another thought. We should give users an option to turn this error reporting off. Especially during the build process, there are some things you don't necessarily want to leak (i.e. secrets etc.).

lforst:
> I would have expected to get distributed tracing when making requests to sentry API as we are sending sentry-trace. Let's discuss why this could happen. Different org maybe?

Yeah that's it. You can have traces across projects but not across orgs afaik.

Lms24:
That looks awesome @vladanpaunovic! The transaction looks just like I had imagined :D 

Anyway, I'm +1 on making a custom client (let's try to have a minimal setup, no or only necessary integrations) and making this opt-out for users

lforst (approved):

PR #38, opened 2022-08-25
feat: Implement `entries` option

### LINKED ISSUE #28: Idea how to implement `entries` option
Look at `isEntry` in `resolveId`. If it's not an entry module, we store it in a map `M`. in `transform` we only inject the `injector` when `id` is not part of `M` - that way we kinda robustly inject only in entry points. Additionally, we can also filter for things the user provides in the `entries` option.

lforst:
Changes the unplugin to only inject the release variable in entrypoint files and provides functionality for the `entries` option we currently have in the webpack plugin.

Resolves #28


Lms24 (approved):
👍

PR #41, opened 2022-08-26
chore(options): add silent

vladanpaunovic:


vladanpaunovic:
We hate this flag, hence this PR is R.I.P.

vladanpaunovic:
We actually like this flag so we are bringing it up with a slight difference

packages/unplugin/src/index.ts
```diff
@@ -94,12 +96,8 @@ const unplugin = createUnplugin<Options>((originalOptions, unpluginMetaContext)
     options.org
   );
 
-  if (options.telemetry) {
-    // eslint-disable-next-line no-console
-    console.log("[Sentry-plugin]", "Sending error and performance telemetry data to Sentry.");
-    // eslint-disable-next-line no-console
-    console.log("[Sentry-Plugin]", "To disable telemetry, set `options.telemetry` to `false`.");
-  }
```
Lms24, line 102: I think we shouldn't get rid of this if condition here. We only want to log this, if Sentry is enabled

packages/unplugin/src/sentry/logger.ts
```diff
@@ -0,0 +1,47 @@
+import { makeSentryClient } from "../sentry/telemetry";
+import { Options } from "../types";
+import { SeverityLevel } from "@sentry/node";
+
+export default function logger(props: Pick<Options, "silent" | "telemetry" | "org">) {
```
Lms24, line 5: Any chance we can pass the hub or client from `index.ts` here so that we don't have to create a new client in this function?

Lms24: If we do this, we can also get rid of the `"telemetry"` property in `props`.

packages/unplugin/src/index.ts
vladanpaunovic: Indeed. However, when creating sentry client inside the logger, we are passing `!!options.telemetry` which dictates whether something will be logged out or not. But you gave me a good idea for a test case!

packages/unplugin/src/sentry/logger.ts
vladanpaunovic: that sounds nice as I too don't really like that we have to re-instantiate client

Lms24 (approved):

PR #42, opened 2022-08-26
feat: Add individual bundler packages

lforst:


packages/eslint-configs/package.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "eslint-config-local",
+  "name": "@sentry-internal/eslint-config",
```
Lms24, line 2: Does this still work? IIRC you said yesterday that the package needs to be named `eslint-config` (?)
Or does it work b/c scope doesn't count?

packages/esbuild-plugin/.babelrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["@babel/env", "@babel/typescript"]
+}
```
Lms24, line 3: Totally optional: 
I think the values here are the same for all projects, right? Can we extract this file into one package?

packages/esbuild-plugin/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.(t|j)sx?$": ["@swc/jest"],
+  },
+};
```
Lms24, line 6: Totally optional:
Any change, we can also extract this to a central package?
We might wanna add a "dev-utils" package (I know this sounds oddly familiar....)

packages/esbuild-plugin/rollup.config.js
```diff
@@ -0,0 +1,35 @@
+import resolve from "@rollup/plugin-node-resolve";
```
Lms24, line 1: Totally optional:
Might also be a candidate for a centralized config?

packages/rollup-plugin/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
Lms24, line 1: Optional: 
Hmmm WDYT about adding something like `packages/*/dist` into the root .gitignore instead?

Lms24 (approved):
🚀 
(Just some minor optional nits but totally happy to revisit this later)

packages/rollup-plugin/.gitignore
lforst: More risky/opaque as to why something is "gitignored". I'll leave it for now.

packages/esbuild-plugin/jest.config.js
lforst: If we notice that having stuff duplicated becomes a problem we can extract this. I usually don't like drying things too early. Same goes for other comments.

PR #43, opened 2022-08-26
feat(options): Add dist

vladanpaunovic:


lforst:
Afaik `dist` also has an influence on how files are uploaded, so this PR doesn't fully implement the `dist` option. Don't know how to go forward here.

vladanpaunovic:
@lforst, I took the description from CLI here https://github.com/getsentry/sentry-cli/blob/master/js/index.d.ts#L114-L118

Tbh, I am not sure i there is more that `dist` does then what this PR. 

@kamilogorek, can you help us what `--dist` actualy does? Is it doing anything more then `myRelease@[DIST_HERE]`?

kamilogorek:
It doesnt change the release name, it updates request form body with `dist` value:

https://github.com/getsentry/sentry-cli/blob/7b8466885d9cfd51aee6fdc041eca9f645026303/src/api.rs#L696-L698

vladanpaunovic:
Ok, I think I got it now. It's actually used for uploading a release file, like here https://docs.sentry.io/api/releases/upload-a-new-organization-release-file/

Lms24:
Closing in favour of #85

PR #34, opened 2022-08-24
ref(cli): Remove Sentry CLI

lforst:
Sleep well.

PR #46, opened 2022-08-26
ref: remove sentryFacade

Lms24:
This PR removes our `SentryFacade` abstraction from the project. As we got rid of Sentry CLI (#34) there is no need anymore to have this additional layer and we can just call the facade's functions directly.

Additionally, the PR introduces the `SentryContext` type which should hold "sentry internal" values/data, like the hub and the parent span. We might wanna add the logger here later (cc @vladanpaunovic). This replaces the arguments we pass down to the release pipeline functions so we can add more stuff w/o having to add yet another param.

### DIFF
--- a/packages/unplugin/src/types.ts
+++ b/packages/unplugin/src/types.ts
@@ -1,4 +1,8 @@
 //TODO: JsDoc for all properties
+
+import { Hub } from "@sentry/hub";
+import { Span } from "@sentry/tracing";
+
 //TODO: compare types w/ webpack plugin (and sentry-cli?)
 export type Options = {
   debugLogging?: boolean;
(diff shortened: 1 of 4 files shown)


packages/unplugin/src/types.ts
```diff
@@ -84,3 +88,12 @@ type IncludeEntry = {
   //TODO: what about the other entries??
 };
 */
+
+/**
+ * Holds data for Sentry internal purposes
+ * (e.g. telemetry and logging)
+ */
+export type SentryContext = {
```
lforst, line 96: `s/SentryContext/BuildContext/`? Just to indicate that this can be used for more stuff than just Sentry SDK objects?

lforst (approved):
nice

PR #47, opened 2022-08-28
feat(repo): Add version bump commands and scripts

Lms24:
This PR adds a yarn script (`bump:version`) to all repo projects. It calles `yarn version` to use yarn for setting a new version to the package.json files. To also bump the top level package.json version, this PR adds a small bash script to first execute the bump in all packages and then set the same version in the main repo project. 

We need something like this to set up releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sentry/sentry-unplugin-root",
-  "version": "0.0.0",
+  "version": "2.0.0-alpha.0",
   "description": "Root of the sentry unplugin monorepo.",
   "repository": "git@github.com:lforst/sentry-unplugin.git",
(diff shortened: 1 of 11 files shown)


lforst:
I don't know why CI is failing [like that](https://github.com/getsentry/hackweek-sentry-unplugin/runs/8059640671?check_suite_focus=true). Could be because `*` in `dependencies` doesn't pick up `-alpha` versions?

Just an idea if that is actually the case: Is there some way we can leave the `@sentry-internal/*` packages on some impossible version like `0.0.0` and only bump the `@sentry/*` packages?

Lms24:
The interesting thing is that for me locally, I can build everything just fine...
will revisit this soon

PR #49, opened 2022-08-30
ref(unplugin): Clean up logger

lforst:
Just some opinionated cleaning up of the logger (reducing API surface, minor naming changes, test cleanup, and so forth).

Some changes are objective though:
- We call `console.log` instead of `console.warn` and `console.error`. Reason is that `.warn` is just an alias for `.error` and `.error` writes to stderr instead of stdout and we don't want that.
- Don't expose `prefix` on logger - it's unnecessary.
- Instead of prefixing with `[Sentry-unplugin]` we're now prefixing with `[sentry-{insert bundler here}-plugin]`.

### DIFF
--- a/packages/unplugin/src/sentry/logger.ts
+++ b/packages/unplugin/src/sentry/logger.ts
@@ -1,15 +1,14 @@
-import { Options } from "../types";
-import { SeverityLevel } from "@sentry/node";
-import { Hub } from "@sentry/node";
-interface LoggerI {
-  options: Pick<Options, "silent" | "org">;
-  hub: Hub;
-}
-export default function logger(props: LoggerI) {
-  const prefix = "[Sentry-unplugin]";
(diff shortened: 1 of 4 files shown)


vladanpaunovic (approved):
Beautiful! Thanks!

packages/unplugin/src/sentry/logger.ts
```diff
@@ -1,43 +1,41 @@
...
       }
 
       addBreadcrumb("info", message);
     },
     warn(message: string) {
-      if (!props.options.silent) {
+      if (!options?.silent) {
         // eslint-disable-next-line no-console
-        console.warn(prefix, message);
+        console.log(`${options.prefix} Warning: ${message}`);
```
vladanpaunovic, line 30: nit:
```suggestion
        console.log(`${options.prefix} Warning! ${message}`);
```

Issue #50, opened 2022-08-31
Roadmap to Sentry Bundler Plugins


lforst:
# Roadmap to Package Release

This issue serves as a roadmap to releasing Sentry bundler plugins for Vite, esbuild, Rollup. Eventually, we'll also replace our Sentry Webpack plugin with this new plugin.

## Necessary for initial release

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-5ed8fed2-765d-4f89-9f34-61425fe21f26

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-e151d2cd-db49-4556-adca-479ebb3ef33b

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-58033726-21b8-4f83-843c-4ab4a4bf3f43

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-51c2de9d-d960-4ecc-99ab-daa6b7f7cd0c

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-8e181a81-eb09-4be9-be43-0cec95f0f3d2

### After the initial release

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50#tasklist-block-883e7144-21d5-4a9c-a00d-9b4e20df0d80

smeubank:
the goal for now is not to replace webpack-plugin right?

lforst:
> the goal for now is not to replace webpack-plugin right?

It is kinda. In intermediary steps we can release packages just for the other bundlers, but goal is to be able to replace the webpack plugin.

PR #53, opened 2022-10-19
feat(publish): Add version bump script

Lms24:
This PR (unfortunately) adds lerna to our repo to use it as a mechanism of bumping all our package versions. We need this for the craft publishing process.

For future readers, here are some reasons for choosing lerna:
- Nx doesn't have a version bump command built in :(
- I tried to use the Nx [semVer](https://www.npmjs.com/package/@jscutlery/semver) plugin but this tool does too much for our purposes
  - it creates git tags (can't disable this)
  - it automatically determines if the version should be a major/minor/patch bump
  - we could only manually override the decision to bump to the next major/minor/patch but we can't just specify a version. 
  -  Because of these limitations and for time reasons we chose not to pursue this approach further
- I tried to use `yarn version` but this would only bump the version of the respective package, but not of the package's dependencies.
  - This caused `yarn install` failures because of dependency missmatches after bumping the versions.

Lerna has a `lerna version` command which lets us set a specific version on our terms. So we're welcoming it back for now until Nx hopefully figures this out. 

Furthermore, the PR adds a pre-release script we need for craft later on

ref #50

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,0 +1,6 @@
+{
+  "version": "0.0.0-alpha.0",
+  "packages": "packages/*",
+  "npmClient": "yarn",
+  "useWorkspaces": true
+}
--- a/package.json
+++ b/package.json
@@ -21,8 +21,9 @@
   "devDependencies": {
     "@nrwl/cli": "14.5.10",
     "@nrwl/workspace": "14.5.10",
-    "nx": "14.5.10",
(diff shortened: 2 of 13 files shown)


lforst (approved):

AbhiPrasad:
Maybe one day we try https://rushjs.io/ - which has this stuff built in.

PR #55, opened 2022-10-20
feat(publish): Add GHA and Craft configs

Lms24:
This adds a few files to set up Craft and our release process.

Draft for testing purposes

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -13,9 +13,5 @@ requireNames:
 targets:
   - name: github
     includeNames: /^sentry-.*.tgz$/
-    # TODO: Remove once we replace the webpack plugin
-    excludeNames: /^sentry-webpack-plugin-*.tgz$/
   - name: npm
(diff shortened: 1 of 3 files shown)


.craft.yml
```diff
@@ -0,0 +1,21 @@
+github:
+  owner: getsentry
+  repo: bundler-plugins
+changelogPolicy: simple
+preReleaseCommand: bash scripts/craft-pre-release.sh
```
Lms24, line 5: Technically, this is an optional property as by default, Craft will look for a script called `scripts/bump-version.sh`. However, I think having this written here explicitly as it makes it clear what this script is doing in the repo and where it is needed.

Lms24:
I'd like to merge this in to get the "Prepare Release" workflow to work properly. There might still be problems with overall craft/GHA config but we kinda need the prepare release WF to show up first s.t. we can start it and provide the inputs it needs.

lforst (approved):

PR #64, opened 2022-10-21
fix(publish): Fix repository and homepage values in root package.json

Lms24:
After our initial package release I noticed that the NPM page of the `@sentry/bundler-plugin-core` package still pointed to the old `getsentry/sentry-unplugin` repo. This PR updates two `package.json`s that still had outdated `repository` and `homepage` values.


lforst (approved):
whoops

Issue #62, opened 2022-10-20
Support `include` and file collection sub-options


Lms24:
`include` is the most important option option for the plugins and we have to support it properly. 
`include` can either be a simple string, an `IncludeEntry` or an array with a mixture of strings and entries. Ideally, we normalize the given user option into an internal option and take it from there. 

Prework:

- [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/66
- [x] Convert options to internal options (https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/69)

Sub-options to support that deal with file collection:
- [x] Collect files to upload (`paths` and `ext` options ) (https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/73)
- [x] `ignore` (#79) 
- [x] `ignoreFile` (#80)

EDIT: moved `include` file modification options to separate issue (#81) 

ref #50

PR #66, opened 2022-10-21
Add all options including JSDoc to `Options` type

Lms24:
This adds all options + JSDoc from the webpack plugin to our `Options` type. 

Some things in this PR we can discuss before merging, e.g. what to do with the `process.env` thing or if we can get rid of the `configFile` option (which I'd vote yes on). 

Others (e.g. `customHeaders` vs `customHeader`), I'd like to tackle in a follow-up PR as it involves modifying more code than just changing the type. 

ref: #62

### DIFF
--- a/packages/bundler-plugin-core/src/types.ts
+++ b/packages/bundler-plugin-core/src/types.ts
@@ -7,9 +7,35 @@ import { createLogger } from "./sentry/logger";
 //TODO: compare types w/ webpack plugin (and sentry-cli?)
 export type Options = {
   /* --- authentication/identification: */
+  /**
+   * The slug of the Sentry organization associated with the app.
+   *
+   * This value can also be specified via `process.env.SENTRY_ORG` (TODO, can it?)
+   */
   org?: string;
+
+  /**
+   * The slug of the Sentry project associated with the app.
+   *
+   * This value can also be specified via `process.env.SENTRY_ORG` (TODO, can it?)
+   */
   project?: string;
+
+  /**
+   * The authentication token to use for all communication with Sentry.
+   * Can be obtained from https://sentry.io/settings/account/api/auth-tokens/.
+   * Required scopes: project:releases (and org:read if setCommits option is used).
+   *
+   * This value an also be specified via `process.env.SENTRY_AUTH_TOKEN` (TODO, can it?).
+   */
   authToken?: string;
+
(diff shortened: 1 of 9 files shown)


lforst (approved):

packages/bundler-plugin-core/src/types.ts
```diff
@@ -1,20 +1,61 @@
...
+   * which automatically detects values for Cordova, Heroku, AWS CodeBuild, CircleCI,
+   * Xcode, and Gradle, and otherwise uses HEAD's commit SHA. (For HEAD option,
+   * requires access to git CLI and for the root directory to be a valid repository).
+   */
   release?: string;
-  // dist: string,
+
+  /**
+   * 	Unique identifier for the distribution, used to further segment your release.
+   *  Usually your build number.
```
lforst, line 56:
```suggestion
   * Unique identifier for the distribution, used to further segment your release.
   * Usually your build number.
```

Lms24:
For future readers:

We decided to make slight deviations from the original Sentry Webpack plugin options:

* Removed `configFile` property as our new bundler plugins do not rely on Sentry CLI anymore. As a consequence, there's no need to specify a `.sentryclirc` file anymore. This means that parameters that were previously set in Sentry CLI which were also relevant for the Webpack plugin need to be set explicitly in the plugin options. 
* Removed implicit passing of plugin parameters via environment variables. Previously, it was possible to specify values implicitly as environment variables, such as `SENTRY_AUTH_TOKEN`. In our new plugin, you'll have to specify these values in the options. Note that this makes certain option fields **explicitly required** now which were previously only implicitly required.
  ```js
  // old config + environment variables were set for authToken, org and project 
  new SentryCliPlugin({
      include: './dist',
    })

  // new config (you can still use env variables but they need to be set explicitly):
  sentryWebpackPlugin({
    include: './dist',
    authToken: process.env.SENTRY_AUTH_TOKEN,
    org: process.env.SENTRY_ORG,
    project: process.env.SENTRY_PROJECT,
  });
  ```

Additional deviations:
* Removed `apiKey` property. This was never directly a part of the webpack plugin but the JS interface of Sentry CLI allowed for this property to be passed in. If you used only the webpack plugin before, this will most likely not affect you.

All of these changes will be written down in a migration guide coming up soon.

PR #67, opened 2022-10-24
fix: Use `writeBundle` hook instead of `buildEnd`

lforst:
Replaces usage of the `buildEnd` hook with usage of the `writeBundle` hook. The `buildEnd` hook only runs before bundles have been written to disk and that is too early for sourcemaps upload - the `writeBundle` hook runs late enough.


Lms24 (approved):
Awesome! 🚀

PR #68, opened 2022-10-24
chore: Add MIGRATION.md guide

Lms24:
This adds a `MIGRATION.md` file to our repo to add breaking changes to. I added the two breaking changes we discussed in #66 as a start. We can iterate on this document and adjust/add changes as we progress towards replacing the webpack plugin.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -4,7 +4,7 @@ This document serves as a migration guide, documenting all breaking changes betw
 
 ## [Unreleased] Upgrading from 1.x to 2.x
 
-Version 2 of `@sentry/webpack-plugin` is a complete rewrite of version 1. Version 2 no longer required `sentry-cli` underneath, which means that the plugin no longer downloads a binary but implements all its functinality natively.
+Version 2 of `@sentry/webpack-plugin` is a complete rewrite of version 1. Version 2 no longer requires `sentry-cli` underneath, meaning the plugin no longer downloads a binary but implements all its functionality natively.
 
 ### Initialization and Required Values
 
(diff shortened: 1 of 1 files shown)


MIGRATION.md
```diff
@@ -0,0 +1,35 @@
+# Migration Guide
+
+This document serves as a migration guide, documenting all breaking changes between major versions of the Sentry bundler plugins.
+
+## [Unreleased] Upgrading from 1.x to 2.x
+
+Version 2 of `@sentry/webpack-plugin` is a complete rewrite of version 1. Version 2 no longer required `sentry-cli` underneath, which means that the plugin no longer downloads a binary but implements all its functinality natively.
```
lforst, line 7:
```suggestion
Version 2 of `@sentry/webpack-plugin` is a complete rewrite of version 1. Version 2 no longer requires `sentry-cli` underneath, meaning the plugin no longer downloads a binary but implements all its functionality natively.
```

MIGRATION.md
```diff
@@ -0,0 +1,35 @@
...
+  include: "./dist",
+  authToken: process.env.SENTRY_AUTH_TOKEN,
+  org: process.env.SENTRY_ORG,
+  project: process.env.SENTRY_PROJECT,
+});
+```
+
+### Removal of Implicit Environment Variable Usage
+
+Version 2 of the Webpack plugin removes the implicit passing of plugin parameters via environment variables. Previously, it was possible to specify values as environment variables, such as SENTRY_AUTH_TOKEN, but to never mention them in the plugin init options. In this version, you'll have to specify these values in the options. Note that this makes certain option fields explicitly required now which were previously only implicitly required (see [Initialization and Required Values](#initialization-and-required-values)).
```
lforst, line 30: l: Do we wanna move this section above "Initialization and Required Values"? That way, readers know what the `process.env.*` stuff in the code snippet is about.

lforst (approved):

PR #73, opened 2022-10-25
feat: Implement `paths` and `ext` options for `include`

lforst:
Ref: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/62

This PR adds the logic to grab files the user configured via `include` and `paths` and applies logic for the correct release artifact prefixes:

- If the user points `paths` to a file, it's artifact path will be its basename prefixed with the `urlPrefix` (eg. `/foo/bar/baz.js` -> `~/baz.js`)
- If the user points `paths` to a directory, the plugin will include all the files recursively included in that directory. The file's artifact paths will be the file path, but the part that the user provided via `paths` is replaced with the `urlPrefix`.

We also filter based on file extensions with the `ext` option.

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
+++ b/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
@@ -91,11 +91,34 @@ export async function uploadSourceMaps(
   options.include.forEach((includeEntry) => {
     includeEntry.paths.forEach((path) => {
       //TODO: handle include property properly
-      files.push(...getFiles(path, includeEntry.ext));
+      files.push(...getFiles(path, includeEntry));
     });
   });
 
   ctx.logger.info(`Found ${files.length} files to upload.`);
+
+  // Check if there would be duplicate artifacts and throw if there are any.
+  const duplicateArtifacts = new Set<string>();
+  const fileSet = new Set<string>();
+  files.forEach((file) => {
+    if (fileSet.has(file.name)) {
+      duplicateArtifacts.add(file.name);
+    } else {
+      fileSet.add(file.name);
+    }
+  });
+  if (duplicateArtifacts.size > 0) {
+    const artifactsList: string[] = [];
+    duplicateArtifacts.forEach((artifact) => {
+      artifactsList.push(`- "${artifact}"`);
+    });
+    ctx.logger.error(
+      `The following artifacts were identified more than once. Use the "urlPrefix" or "urlSuffix" options to tell them apart or adjust your "include" and "ignore" settings.\n${artifactsList.join(
+        "\n"
+      )}`
+    );
+    throw new Error();
+  }
 
   return Promise.all(
     files.map((file) =>
--- a/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
+++ b/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
@@ -1,38 +1,66 @@
 import path from "path";
-import fs from "fs";
+import fs, { Stats } from "fs";
+import glob from "glob";
+import { InternalIncludeEntry } from "../options-mapping";
 
 export type FileRecord = {
   name: string;
   content: string;
 };
 
-export function getFiles(path: string, allowedExtensions: string[]): FileRecord[] {
-  const includedFiles = getAllIncludedFileNames(path, allowedExtensions, []);
+export function getFiles(fpath: string, includeEntry: InternalIncludeEntry): FileRecord[] {
+  let fileStat: Stats;
+  const p = path.isAbsolute(fpath) ? fpath : path.resolve(process.cwd(), fpath);
+  try {
+    fileStat = fs.statSync(p);
+  } catch (e) {
+    return [];
+  }
 
-  return includedFiles.map((filename) => {
-    const content = fs.readFileSync(filename, { encoding: "utf-8" });
-    return { name: "~" + filename.replace(new RegExp(`^${path}`), ""), content };
+  let files: {
+    absolutePath: string;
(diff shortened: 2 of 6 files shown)


packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -91,12 +91,35 @@ export async function uploadSourceMaps(
   options.include.forEach((includeEntry) => {
     includeEntry.paths.forEach((path) => {
       //TODO: handle include property properly
```
Lms24, line 93: Do we still need this TODO?

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -91,12 +91,35 @@ export async function uploadSourceMaps(
...
+  if (duplicateArtifacts.size > 0) {
+    const artifactsList: string[] = [];
+    duplicateArtifacts.forEach((artifact) => {
+      artifactsList.push(`- "${artifact}"`);
+    });
+    ctx.logger.error(
+      `The following artifacts were identified more than once. Use the "urlPrefix" or "urlSuffix" options to tell them apart or adjust your "include" and "ignore" settings.\n${artifactsList.join(
+        "\n"
+      )}`
+    );
```
Lms24, line 119: super-l (sorrry, feel free to ignore): We could probably get away with a `duplicateArtifacts.reduce` call here to directly create one string, right? Would spare us from joining the list in the error message.

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -1,38 +1,66 @@
...
 export type FileRecord = {
   name: string;
   content: string;
 };
 
-export function getFiles(path: string, allowedExtensions: string[]): FileRecord[] {
-  const includedFiles = getAllIncludedFileNames(path, allowedExtensions, []);
+export function getFiles(fpath: string, includeEntry: InternalIncludeEntry): FileRecord[] {
+  let fileStat: Stats;
+  const p = path.isAbsolute(fpath) ? fpath : path.resolve(process.cwd(), fpath);
```
Lms24, line 13: l: Let's try to give this a more expressive name for self-doc reasons

Suggestion (but feel free to call it something else):
```suggestion
  const absolutePath = path.isAbsolute(fpath) ? fpath : path.resolve(process.cwd(), fpath);
```

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -1,38 +1,66 @@
...
+  }[];
+
+  if (fileStat.isFile()) {
+    files = [{ absolutePath: p, relativePath: path.basename(p) }];
+  } else if (fileStat.isDirectory()) {
+    files = glob
+      .sync(path.join(p, "**"), {
+        nodir: true,
+        absolute: true,
+      })
```
Lms24, line 33: That's a great way of getting rid of our own recursive file search :D

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -1,38 +1,66 @@
...
+      .map((globPath) => ({ absolutePath: globPath, relativePath: globPath.slice(p.length + 1) }));
```
Lms24, line 34: Q: What does `globPath.slice(p.length + 1)` return?

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -1,38 +1,66 @@
...
+  } else {
+    return [];
+  }
+
+  const dotPrefixedAllowedExtensions = includeEntry.ext.map(
+    (extension) => `.${extension.replace(/^\./, "")}`
```
Lms24, line 40: Q: Do we do this to sanitize user input without a leading dot?

If yes, should we move this to the `normalizeUserOptions` function?

Lms24 (approved):
Looks good to me! 
Just a couple of super-l nits and questions but nothing blocking.

Let's not forget to add a note to `MIGRATION.md` that we're not supporting globs anymore (doesn't have to happen in this PR though)

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
Lms24: Ahh nevermind, I figured it out. This basically subtracts the root path `p` from the path we get from glob. Nice!

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
lforst: Nope thank you!

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
lforst: `duplicateArtifats` is a `Set` and it doesn't have the reduce method. I would do it that way if I could :(

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
lforst: Yeah that was pure laziness during development. Wanted to change that variable before pushing the PR - thanks for pointing out!

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
lforst: Good point. I moved it to the normalization function.

PR #74, opened 2022-10-27
ref: Clean up `include` logic

lforst:
Implements feedback from #73


Lms24 (approved):
Thanks for addressing my feedback!

PR #77, opened 2022-10-28
 test(CI): Add E2E and split test jobs

Lms24:
This PR splits up our tests in CI and adds an E2E test job. We now have three jobs for testing:

* Unit tests
* Integration tests
* E2E tests

New testing commands for yarn:
* `yarn test`
  * top level: executes all tests **except** for E2E tests
  * package level: executes the tests of the package (also in E2E tests package)
* `yarn test:all`
  * top level: executes all tests, **including** E2E tests
  * package level: N/A

In the future we might want to look into build caching to leverage nx' cache capabilities but for now, I'll work on more pressing stuff. 

This change also includes a newly added secret to the repo for the Sentry auth token. Furthermore, I temporarily removed the "required" checks for the no longer existing "Tests" job. Will mark the 3 jobs as required once this CI change is merged and available in `main`.


nx.json
```diff
@@ -8,7 +8,8 @@
     "default": {
       "runner": "nx/tasks-runners/default",
       "options": {
-        "cacheableOperations": ["build", "lint", "test"]
+        "cacheableOperations": ["build", "lint", "test"],
+        "parallel": 10
```
Lms24, line 12: At least locally, this speeds things up a little for me.

package.json
```diff
@@ -17,6 +17,9 @@
     "clean": "nx run-many --target=clean --all",
     "clean:all": "nx run-many --target=clean:all --all && yarn",
     "test": "nx run-many --target=test --all",
+    "test:unit": "nx run-many --target=test --all --exclude=@sentry-internal/integration-tests,@sentry-internal/bundler-plugin-e2e-tests",
+    "test:integration": "nx run @sentry-internal/integration-tests:test",
+    "test:e2e": "nx run @sentry-internal/bundler-plugin-e2e-tests:test",
```
lforst, line 22: I like this

packages/e2e-tests/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:e2e": "run-s test:setup test:jest",
+    "test": "run-s test:setup test:jest",
```
lforst, line 7: If we have the `test` command here, won't running `yarn test` in the workspace root try to run the E2E tests? I am worried that this is gonna be annoying if you don't have an auth token configured.

lforst (approved):

packages/e2e-tests/package.json
Lms24: Good point, thanks. My main motivation for renaming this to `test` was to still make `yarn test` execute all tests in the repo. But I agree that this is probably annoying for people to set up. I'll revert this change and add a top-level `test:all` command that includes E2E tests.

Lms24: Update: I opted for another solution because what I described above required a few not so nice hacks and I feel like `test` should always be the main test execution command. Here's what I'm gonna do:
* `yarn test`
  * top level: executes all tests **except** for E2E tests
  * package level: executes the tests of the package (also in E2E tests package)
* `yarn test:all`
  * top level: executes all tests, **including** E2E tests
  * package level: N/A

PR #79, opened 2022-10-28
feat(core): Add `ignore` option

Lms24:
This PR adds the `ignore` option implementation to the plugin core. We're using `glob` for recursively searching for files which, as luck would have it, also supports an `ignore` option (h/t @lforst for finding this out). 

We don't have to do a lot to get it working properly other than converting the ignore entries from our users to absolute paths. I tried passing on relative paths but it wouldn't work as expected when trying to replicate how users (and our [documentation](https://github.com/getsentry/sentry-webpack-plugin#optionsinclude)) [use the `ignore` option](https://cs.github.com/?scopeName=All+repos&scope=&q=%22new+sentrycliplugin%28%22+ignore). 

Side note: I decided to split this and the `ignoreFile` options up into separate PRs as the latter one requires some additional work and probably another dependency. Currently figuring that out.

ref: #62

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
+++ b/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
@@ -28,7 +28,9 @@ export function getFiles(includePath: string, includeEntry: InternalIncludeEntry
   if (fileStat.isFile()) {
     files = [{ absolutePath, relativePath: path.basename(absolutePath) }];
   } else if (fileStat.isDirectory()) {
-    const absoluteIgnores = includeEntry.ignore.map((e) => path.join(absolutePath, e));
+    const absoluteIgnores = includeEntry.ignore.map((ignoreEntry) =>
+      path.join(process.cwd(), ignoreEntry)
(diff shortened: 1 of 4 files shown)


packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -28,10 +28,13 @@ export function getFiles(includePath: string, includeEntry: InternalIncludeEntry
   if (fileStat.isFile()) {
     files = [{ absolutePath, relativePath: path.basename(absolutePath) }];
   } else if (fileStat.isDirectory()) {
+    const absoluteIgnores = includeEntry.ignore.map((e) => path.join(absolutePath, e));
```
Lms24, line 31: Ok, so thinking about this for another round, I might have to swap this to path.join(process.cwd(), ignoreEntry) because that makes more sense w.r.t to what's documented. (also the "node_modules" default only makes sense in that context).

lforst (approved):

PR #80, opened 2022-10-31
feat(core): Add `ignoreFile` option

Lms24:
This PR implements the `ignoreFile` option and rewrites the `ignore` option behaviour. Previously, we used `glob`'s `ignore` option to pass an array of ignore rules to the library. This meant we'd get back a set of filtered files where ignore rules were already applied. However, `glob` does not support negated ignore rules, meaning ignore patterns like "ignore everything except file x" wouldn't work. 

With this PR, we introduce a new library to the project: [ignore](https://www.npmjs.com/package/ignore). It lets us add rules to a rule checker (regardless of if these rules come from a file or an array). We now first collect all files with `glob` and subsequently filter out the ones that are ignored by the specified ignore rules. This rule checker supports all rules as specified in `.gitignore` specs. Therefore, it also supports negated ignore rules.

ref: #62

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
+++ b/packages/bundler-plugin-core/src/sentry/sourcemaps.ts
@@ -19,7 +19,7 @@ type FileNameRecord = {
 };
 
 export function getFiles(includePath: string, includeEntry: InternalIncludeEntry): FileRecord[] {
-  // Start with getting all files for the given includePath. If there are non, we can bail at this point.
+  // Start with getting all (unfiltered) files for the given includePath.
   const files = collectAllFiles(includePath);
   if (!files.length) {
     return [];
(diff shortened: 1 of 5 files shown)


packages/bundler-plugin-core/src/sentry/sourcemaps.ts
```diff
@@ -19,52 +61,46 @@ export function getFiles(includePath: string, includeEntry: InternalIncludeEntry
     return [];
   }
 
-  let files: {
-    absolutePath: string;
-    // Contains relative paths without leading `.` (e.g. "foo/bar.js" or "asdf\\a\\b.js.map")
-    relativePath: string;
-  }[];
+  let files: FileNameRecord[];
```
lforst, line 64: Is there a reason we define a `let` here instead of just returning directly in the if-else?

lforst (approved):

packages/bundler-plugin-core/src/sentry/sourcemaps.ts
Lms24: Nope, no particular reason. Changed it in [9253137](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/80/commits/9253137ab8ccde3a2602eee782f3054d3d01b212)

PR #84, opened 2022-11-03
ref(core): Change customHeaders option to customHeader

### LINKED ISSUE #83: Adjust `customHeader` option
To go with feature parity to the webpack plugin, we should rename the option we currently call `customHeaders` to `customHeader` and let users pass a string in the format `[header-name]: [header-value]`.

Lms24:
Rename `customHeaders` to `customHeader` and change user-facing data type to string. This change aligns the option with the Sentry Webpack Plugin.

Note: For internal usage, I left it as `Record<string, string>` because it makes it easier for us to work with. We might consider merging other Http headers into this property in the future. For instance authToken and UserAgent. Then we'd have one internal option (e.g. `headers`) that we could populate during the internal option conversion. 

fixes #83

### DIFF
--- a/packages/bundler-plugin-core/src/options-mapping.ts
+++ b/packages/bundler-plugin-core/src/options-mapping.ts
@@ -12,7 +12,6 @@ type RequiredInternalOptions = Required<
     | "finalize"
     | "validate"
     | "vcsRemote"
-    | "customHeaders"
     | "dryRun"
     | "debug"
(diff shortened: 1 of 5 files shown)


lforst (approved):

Lms24:
I'm gonna go ahead and merge this anyway because I'd like to stick with our internal options, even if we re-add Sentry-CLI

Issue #85, opened 2022-11-03
Re-introduce Sentry CLI


Lms24:
After realizing that achieving feature parity with the Sentry WebPack Plugin and at the same time replacing Sentry CLI underneath would be an enourmous amount of work, we decided to prioritize feature parity. Consequently, we're re-introducing Sentry CLI.

Tasks:
* [x] Add the CLI package as a dependency and create a CLI instance https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/86
* [x] #87 
* [x] #88
* [x] Revisit create release and other partially implemented pipeline stages
  * https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/90 
  * https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/92
  * https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/93

ref: #50

PR #87, opened 2022-11-04
ref(core): Replace release detection with Sentry CLI release detection

Lms24:
This PR marks the first step of re-adding CLI functionality. It removes our previous release detection code (RIP) and replaces it with CLI functionality. Doing this to ensure feature parity with the CLI, we can revisit this if we have time.

ref #85


packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -81,7 +80,7 @@ export function normalizeUserOptions(userOptions: UserOptions): InternalOptions
     project: userOptions.project,
     authToken: userOptions.authToken,
     url: userOptions.url ?? "https://sentry.io/",
-    release: userOptions.release ?? detectRelease(),
+    release: userOptions.release ?? "",
```
Lms24, line 83: This seems odd but I want to keep `release` a required property, hence I'm assigning an empty string.

packages/bundler-plugin-core/src/index.ts
```diff
@@ -116,13 +119,17 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
     /**
      * Responsible for starting the plugin execution transaction and the release injection span
      */
-    buildStart() {
+    async buildStart() {
+      if (!internalOptions.release) {
+        internalOptions.release = await cli.releases.proposeVersion();
+      }
```
Lms24, line 125: In case we don't have a release (i.e. an empty sting, we use the CLI to detect it automatically)

packages/bundler-plugin-core/src/options-mapping.ts
lforst: Would it hurt us to make it an optional property? If it is actually optional now, we should define it as such. (Same probably down the line for project, authToken etc. because they could be defined via config/env-vars now right?)

lforst (approved):

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: `release` is a special case here because we can't easily await the `cli.releases.proposeVersion` promise in the top-level function we pass to `createUnplugin`. Hence I moved it to the `beforeBuild` hook

Lms24: > because they could be defined via config/env-vars now right?

@lforst especially the config file is a valid point. Haven't thought about this when making the change here. So this means that, unless we want to read and parse the config file during normalization (which I don't want to do unless necessary), we have to refactor our internal options conversion to actually make these parameters (releaes, authtoken, etc) really optional... 

I'm gonna merge this PR to get some of the quick CLI replacements in and then tackle the options again. In the end this will change this line above

PR #88, opened 2022-11-04
ref(core): Use Sentry CLI for sourcemaps upload

Lms24:
To be merged after #87

This replaces our own WIP sourcemaps upload logic with Sentry CLI's implementation.  

ref #85


lforst (approved):
Rip all that logic :D

Lms24:
> Rip all that logic :D

Good point, made me realize we can get rid of the entire `sourcemaps.ts` file 🥲

PR #92, opened 2022-11-07
ref(core): Use Sentry CLI for artifact deletion

Lms24:
Replace previous custom implementation with Sentry CLI logic analogously to the [webpack plugin](https://github.com/getsentry/sentry-webpack-plugin/blob/137503f3ac6fe423b16c5c50379859c86e689017/src/index.js#L487-L490)

ref: #85

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
+++ b/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
@@ -8,7 +8,7 @@
 
 import { InternalOptions } from "../options-mapping";
 import { BuildContext } from "../types";
-import { createRelease, deleteAllReleaseArtifacts } from "./api";
+import { createRelease } from "./api";
 import { addSpanToTransaction } from "./telemetry";
 
 export async function createNewRelease(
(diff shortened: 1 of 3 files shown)


packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -86,40 +86,31 @@ export async function finalizeRelease(options: InternalOptions, ctx: BuildContex
...
-export async function cleanArtifacts(options: InternalOptions, ctx: BuildContext): Promise<string> {
+export async function cleanArtifacts(options: InternalOptions, ctx: BuildContext): Promise<void> {
   const span = addSpanToTransaction(ctx, "function.plugin.clean_artifacts");
 
   if (options.cleanArtifacts) {
     // TODO: pull these checks out of here and simplify them
     if (options.authToken === undefined) {
       ctx.logger.warn('Missing "authToken" option. Will not clean existing artifacts.');
-      return Promise.resolve("nothing to do here");
+      return;
```
AbhiPrasad, line 96: l: really feels like this should be abstracted out - perhaps into an array that gets iterated on (we can then even validate this with types)

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -86,40 +86,31 @@ export async function finalizeRelease(options: InternalOptions, ctx: BuildContex
...
-    await deleteAllReleaseArtifacts({
-      authToken: options.authToken,
-      org: options.org,
-      release: options.release,
-      sentryUrl: options.url,
-      project: options.project,
-      sentryHub: ctx.hub,
-      customHeader: options.customHeader,
-    });
+    await ctx.cli.releases.execute(["releases", "files", options.release, "delete", "--all"], true);
```
AbhiPrasad, line 108: surprised this doesn't have a return value

AbhiPrasad (approved):

PR #93, opened 2022-11-08
ref(core): Use Sentry CLI to create new release

Lms24:
This replaces our custom implementation of the release creation API call with a call to Sentry CLI to do the job

ref: #85


mydea (approved):
LGTM!

PR #96, opened 2022-11-08
feat(core): Add `setCommits`

### LINKED ISSUE #94: Add `setCommit` command

Lms24:
This PR adds the `setCommits` option implementation, analogously to what we have in the [webpack plugin](https://github.com/getsentry/sentry-webpack-plugin/blob/137503f3ac6fe423b16c5c50379859c86e689017/src/index.js#L495-L516).

Also, this adds a few tests to check that this step in our release pipeline is working as expected. I'll make similar tests to the other steps in future PRs.

fixes #94

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -287,7 +287,7 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
       createNewRelease(internalOptions, ctx)
         .then(() => cleanArtifacts(internalOptions, ctx))
         .then(() => uploadSourceMaps(internalOptions, ctx))
-        .then(() => setCommits(ctx)) // this is a noop for now
+        .then(() => setCommits(internalOptions, ctx)) // this is a noop for now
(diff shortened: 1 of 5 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -287,7 +287,7 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
       createNewRelease(internalOptions, ctx)
         .then(() => cleanArtifacts(internalOptions, ctx))
         .then(() => uploadSourceMaps(internalOptions, ctx))
-        .then(() => setCommits(ctx)) // this is a noop for now
+        .then(() => setCommits(internalOptions, ctx)) // this is a noop for now
```
mydea, line 290: I guess we can remove the comment here now? ;)

mydea (approved):

packages/bundler-plugin-core/src/index.ts
Lms24: Indeed 😅

PR #97, opened 2022-11-08
feat(core): Add `deploy` command

### LINKED ISSUE #95: Add `deploy` command
Add the `deploy` option (and the corrseponding command/step in the release pipeline).

ref #50

Lms24:
This PR adds the implementation of the `deploy` option and its step in the release pipeline. The implementation is mostly taken from the [webpack plugin](https://github.com/getsentry/sentry-webpack-plugin/blob/137503f3ac6fe423b16c5c50379859c86e689017/src/index.js#L524-L536) and relies on Sentry CLI.

fixes #95

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
+++ b/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
@@ -127,12 +127,29 @@ export async function setCommits(options: InternalOptions, ctx: BuildContext): P
   span?.finish();
 }
 
-export async function addDeploy(
-  /* version: string, */
-  ctx: BuildContext
-): Promise<string> {
-  const span = addSpanToTransaction(ctx, "function.plugin.add_deploy");
+export async function addDeploy(options: InternalOptions, ctx: BuildContext): Promise<void> {
+  const span = addSpanToTransaction(ctx, "function.plugin.deploy");
+
+  if (options.deploy) {
+    const { env, started, finished, time, name, url } = options.deploy;
+
+    if (env) {
+      await ctx.cli.releases.newDeploy(options.release, {
+        env,
+        started,
+        finished,
+        time,
+        name,
+        url,
+      });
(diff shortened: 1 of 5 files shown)


packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -127,12 +127,29 @@ export async function setCommits(options: InternalOptions, ctx: BuildContext): P
   span?.finish();
 }
 
-export async function addDeploy(
-  /* version: string, */
-  ctx: BuildContext
-): Promise<string> {
-  const span = addSpanToTransaction(ctx, "function.plugin.add_deploy");
+export async function addDeploy(options: InternalOptions, ctx: BuildContext): Promise<void> {
+  const span = addSpanToTransaction(ctx, "function.plugin.deploy");
```
mydea, line 131: l: I guess this is just taken from the previous/current implementation, but does it make sense to create this span even when this is disabled, or would it not make more sense to skip creating this span in that case? Just out of interest :)

mydea (approved):
LGTM! 🚀

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
Lms24: I was also already thinking about this (for all optional steps such as `cleanArtifacts`, `addDeploy` or `setCommits`). I think we have two options: 
* skip the span
* set a different span status (e.g. `cancelled`) if the option is not selected

I've extracted #99 from #50 to track everything we need to revisit in terms of the `telemetry` option

mydea: 👍 personally I'd go with skipping them, as they are basically 0-length spans with little information, and you can infer by their absence/presence if they have been skipped or not. but setting a span status is for sure also OK :)

PR #102, opened 2022-11-09
feat(core): Add `SENTRY_RELEASES` variable during release injection

Lms24:
In our initial implementation of release injection, we did not inject the `SENTRY_RELEASES` global variable which was added to the webpack plugin in https://github.com/getsentry/sentry-webpack-plugin/pull/307. 

Although it is unlikely that people are actually using this feature, it doesn't hurt us a lot if we implement it to prevent breakage for anyone who might manually be relying on this variable (e.g. MFE projects). 

EDIT: We decided to make this behaviour opt-in by introducing a new `injectReleasesMap` option. 

---

A little more context:

The idea of this variable is to have a map of projects that point to the same release. This seems to be a request for module federation and MFEs where it seems that multiple projects want to access the same release.

This original PR was merged, however it seems like our SDK never reads `SENTRY_RELEASES`. The PR author added this in a comment: 

> If this is accepted, we can maybe open another PR on @sentry/browser so that the default release is read from SENTRY_RELEASES and retire SENTRY_RELEASE

So what I'm thinking is that this just never happened. This however still means that people could actually be using this feature by manually accessing `global.SENTRY_RELEASES` and setting the version based on that.

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -13,7 +13,7 @@ import "@sentry/tracing";
 import { addSpanToTransaction, captureMinimalError, makeSentryClient } from "./sentry/telemetry";
 import { Span, Transaction } from "@sentry/types";
 import { createLogger } from "./sentry/logger";
-import { normalizeUserOptions } from "./options-mapping";
+import { InternalOptions, normalizeUserOptions } from "./options-mapping";
 import { getSentryCli } from "./sentry/cli";
 
 // We prefix the polyfill id with \0 to tell other plugins not to try to load or transform it.
@@ -185,7 +185,11 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
       });
 
       if (id === RELEASE_INJECTOR_ID) {
-        return generateGlobalInjectorCode({ release: internalOptions.release });
+        return generateGlobalInjectorCode({
+          release: internalOptions.release,
+          org: internalOptions.org,
+          project: internalOptions.project,
+        });
       } else {
         return undefined;
       }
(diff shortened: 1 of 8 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -333,6 +345,16 @@ function generateGlobalInjectorCode({ release }: { release: string }) {
             {};
 
     _global.SENTRY_RELEASE={id:"${release}"};`;
+
+  if (project) {
+    const key = org ? `${project}@${org}` : project;
+    code += `
+      _global.SENTRY_RELEASES=_global.SENTRY_RELEASES || {};
+      _global.SENTRY_RELEASES["${key}"]={id:"${release}"};
```
AbhiPrasad, line 353: l: can we make this hidden behind a flag? I'd rather not inject things unnecessarily. I recognize this is a breaking change, but given this is undocumented I'm fine making the change.

Lms24: I think that's fair and it would not bloat the bundle unnecessarily for all 99% of users. Will do.

mydea: I agree, is this something that people could have to opt-in to? If that's easy to do, I would say let people opt in (then, in a hypothetical world where we have a better understanding of which options are set by users, we can take better decisions in the future about this functionality). Seems like it is something pretty specific to want.

Lms24: [a3ca32f](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/102/commits/a3ca32fbb9ffc185e91940735469fbd58de768e4)

mydea (approved):

AbhiPrasad (approved):

Issue #91, opened 2022-11-07
Refactor Internal Options


Lms24:
Re-Introducing Sentry CLI has implications on how we convert user options to internal options. We need to make adjustments:

* [x] `customHeader` should just stay a string (https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/98)
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/101
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/103
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/104
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/105

ref: #50

Lms24:
test

PR #103, opened 2022-11-09
ref(core): Adjust requiredness of user and internal option

Lms24:
This PR adjusts some user-facing and internal option from being strictly required to being optional. The reason for this change is that with the introduction of Sentry CLI, we can (and for feature parity should) let users either use env variables but also they can specify some values via a `.sentryclirc` config file. 

Note that for release injection to work properly, users have to specify `project`, `org` and `release` either via the options or via env variables. The reason is that the config file which could also hold these options is only later used by Sentry CLI. This behaviour is identical to what we do in the webpack plugin.

ref #91


packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -52,57 +57,84 @@ export type InternalIncludeEntry = RequiredInternalIncludeEntry &
   };
 
 export function normalizeUserOptions(userOptions: UserOptions): InternalOptions {
-  let entries: (string | RegExp)[] | ((filePath: string) => boolean) | undefined;
-  if (userOptions.entries === undefined) {
-    entries = undefined;
-  } else if (typeof userOptions.entries === "function" || Array.isArray(userOptions.entries)) {
-    entries = userOptions.entries;
+  const entries = normalizeEntries(userOptions.entries);
+  const include = normalizeInclude(userOptions);
```
lobsterkatie, line 61: L: When I first saw this, my instinct was to ask if there was reason for passing just `entries` to `normalizeEntries` but passing the full `userOptions` object to `normalizeInclude`. Having dug into the code as far as `normalizeIncludeEntry`, I can now see why, but might be worth a comment.

L: Also, is there a reason to do these separately, rather than inlining them in the return value, now that they're just function calls?

packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -52,57 +57,84 @@ export type InternalIncludeEntry = RequiredInternalIncludeEntry &
...
+
+  return {
+    // Strictly required options
+    include,
+
+    // These options must be set b/c we need them for release injection
+    // They can also be set as environment variables. Technically, they
+    // could be set in the config file but this would be too late for
+    // release injection
+    org: userOptions.org ?? process.env["SENTRY_ORG"],
```
lobsterkatie, line 71: H: Are we sure that `??` works with all versions of Node someone might be using? Is there a reason not to use `||` here, given that there are no valid falsy values possible here?

(I see that later on we're using `??` for booleans, where clearly `||` won't cut it, but it'd be simple to polyfill with something like 
```js
const useOptionOrDefault /* or choose some better name */ = (
  optionValue,
  defaultValue
) =>
  optionValue !== undefined && optionValue !== null
    ? optionValue
    : defaultValue;
```
.)

packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -52,57 +57,84 @@ export type InternalIncludeEntry = RequiredInternalIncludeEntry &
...
+    return undefined;
+  } else if (typeof userEntries === "function" || Array.isArray(userEntries)) {
+    return userEntries;
   } else {
-    entries = [userOptions.entries];
+    return [userEntries];
   }
+}
+
+function normalizeInclude(userOptions: UserOptions): InternalIncludeEntry[] {
```
lobsterkatie, line 117: L: It might be nice to mention in a docstring here the different formats `include` can take.

packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -52,57 +57,84 @@ export type InternalIncludeEntry = RequiredInternalIncludeEntry &
...
     });
   } else {
-    userInclude = [userOptions.include];
+    userInclude = [rawUserInclude];
   }
 
-  const include = userInclude.map((userIncludeEntry) =>
+  return userInclude.map((userIncludeEntry) =>
     normalizeIncludeEntry(userOptions, userIncludeEntry)
   );
```
lobsterkatie, line 137: M: If you were to use a helper like [the SDK's `arrayify` function](https://github.com/getsentry/sentry-javascript/blob/c59c70bfc636de9bd7512f37e9644db1947a5563/packages/utils/src/misc.ts#L203-L211), you could simplify this a lot, I think, to basically just be something like

```js
return arrayify(userOptions.include)
  .map((potentialIncludeEntry) =>
    typeof potentialUserIncludeEntry === "string"
      ? convertIncludePathToIncludeEntry(potentialIncludeEntry)
      : potentialIncludeEntry
  )
  .map((userIncludeEntry) =>
    normalizeIncludeEntry(userOptions, userIncludeEntry)
  );
```

packages/bundler-plugin-core/src/types.ts
```diff
@@ -12,30 +12,32 @@ export type Options = Omit<IncludeEntry, "paths"> & {
...
+   * This value can also be set via the `SENTRY_AUTH_TOKEN` env variable
    */
-  authToken: string;
+  authToken?: string;
 
   /**
    * The base URL of your Sentry instance. Use this if you are using a self-hosted
    * or Sentry instance other than sentry.io.
    *
+   * This value can also be set via the `SENTRY_URL` env variable.s
```
lobsterkatie, line 39: These are good additions!

packages/bundler-plugin-core/src/types.ts
```diff
@@ -93,13 +97,17 @@ export type Options = Omit<IncludeEntry, "paths"> & {
...
+   *
    * Defaults to 'origin'.
    */
   vcsRemote?: string;
 
   /**
    * A header added to every outgoing network request.
    * The format should be `header-key: header-value`.
+   *
+   * This value can also be set via the `CUSTOM_HEADER` env variable.
```
lobsterkatie, line 110: M: Is there a reason why this isn't `SENTRY_CUSTOM_HEADER`? Also, can you set more than one header here?

lobsterkatie (commented):
All makes sense, though can you explain why reading from the file comes too late? Could you just do that earlier? (Not that that would change what you're doing here necessarily - I think the env variable option is good. More for my own understanding.)

lobsterkatie:
FYI: Hope you don't mind, but I edited the title and PR description to pluralize "options," since I found the singular confusing. (I kept thinking, 'wait, which option?') 

Also, L: You might consider being more specific in the PR title about which direction the adjustment is going (something like `ref(core): Make various user and internal options optional`, for example).

Lms24:
>can you explain why reading from the file comes too late? Could you just do that earlier? 

The reason is that we (and with that I mean the new bundler plugin but also the old webpack plugin) don't do anything with the config file. We just pass through the `configFile` option to Sentry CLI which only then reads it and uses the values specified in there. Release injection doesn't use Sentry CLI for anything other than automatically detecting a release name if none is specified via options or an env variable  ([which does not seem to check env variables or the config file](https://github.com/getsentry/sentry-cli/blob/668d142c1a808f1e695ca11a22663014e4a5d2c0/src/utils/releases.rs#L87)). So the way I see this, there is no way how values from the config file could be accessed for release injection. Unless of course, we say we go beyond feature parity and parse the file ourselves (which would be scope creep).

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: [52564d9](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/103/commits/52564d965589053396952d23d85460923a2087d1)

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: Good catch, thanks! Node 10 (which we're targeting as the min. supported version) [doesn't support](https://node.green/#ES2020-features--nullish-coalescing-operator-----) the nullish coalescing operator. 

It's also a behaviour change for the string values but I think we can justify falling back to  the env/default values if users provided `""`, WDYT? Regardless, I'm wondering if we simply should use the `useOptionOrDefault` option in our source code or polyfill it while transpiling. Our build setup uses rollup with Babel for transpilation. There surely is a way to do this but I have to look into it yet. Any preferences?

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: [85ce2a8](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/103/commits/85ce2a8b507e4245dcd995cf816b8c9275f26c23)

packages/bundler-plugin-core/src/types.ts
Lms24: I was also weirded out by this inconsistent naming but the env variable is called `CUSTOM_HEADER` in Sentry CLI: https://docs.sentry.io/product/cli/configuration/#configuration-values

So we can't really change that as long as we want to maintain compatibility and I'd vote that just for this purpose introducing a new variable in the plugin which differs from the CLI causes more confusion than it brings value

Lms24: Judging by the CLI documentation,

> The header that will be added to every outgoing request in key:value format

I don't think you could add more than one. Although I'm not sure what the CLI would do with a value like `key1:value1\nkey2:value2`.

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: Much cleaner, thanks for the suggestion! 
[081d1de](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/103/commits/081d1de7c5991eb2e8970edc2bd5dd472c72096e)

packages/bundler-plugin-core/src/options-mapping.ts
Lms24: Okay, I should have taken a look at our output: We already transpile the operator away. Here's how the output for the `orgs` option above looks:

```
org: (_userOptions$org = userOptions.org) !== null && _userOptions$org !== void 0 ? _userOptions$org : process.env["SENTRY_ORG"],
```

mydea (approved):
l: As discussed offline, I think it is OK to _not_ handle the env variables here and just leave this to CLI - and pass these through as undefined. This way we do not need to duplicate default etc. logic. It's good to still document which env vars affect these, though 👍 other than this, looks good!

packages/bundler-plugin-core/src/options-mapping.ts
lobsterkatie: Ah, right. Didn't know how low you were targeting your down-compiling.

Seeing that output is a good reminder, though, that we should use `||` where we can for anything that'll be used in the browser - way fewer bytes than the polyfill if we don't actually need it.

packages/bundler-plugin-core/src/types.ts
lobsterkatie: To avoid future deprecation possibility, my gut would be to a) make both work for now, and b) only document that one with `SENTRY` in the name. That way, if/when you do ever move away from using `sentry-cli`, you're not stuck with its legacy weird choices.

UPDATE: We weren't the only ones to have this idea: https://github.com/getsentry/sentry-cli/blob/a08246db0f52438b0c32886148502f6be2591db4/CHANGELOG.md?plain=1#L221

So if we go to `sentry-cli` 2.0, this will be a moot point.

Lms24: I extracted this into an issue (#106). Would you mind giving it a read @lobsterkatie?

Issue #106, opened 2022-11-11
Replace env variable `CUSTOM_HEADER` with `SENTRY_HEADER` 


Lms24:
Extracted from https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/103#discussion_r1019575661:

The `customHeader` option can also be set via the `CUSTOM_HEADER` env variable. This env variable's name is inconsistent as we don't prefix it with `SENTRY_` like with all other env variables. Sentry CLI v1 expects exactly this header.  Sentry CLI v2 expects the `SENTRY_HEADER` env variable (see [changelog](https://github.com/getsentry/sentry-cli/blob/a08246db0f52438b0c32886148502f6be2591db4/CHANGELOG.md?plain=1#L221)) which replaces `CUSTOM_HEADER`.

To already be ahead of a breaking change when upgrading to CLI v2, we can default the JSDoc of the bundler plugin to `SENTRY_HEADER` and internally, when converting to internal options, set the `customHeader` variable to the env variables' value:

```js
customHeader: userOptions.customHeader ?? process.env["SENTRY_HEADER"] ?? process.env["CUSTOM_HEADER"]
``` 

Interestingly, `SENTRY_HEADER` isn't documented in the [CLI configuration docs](https://docs.sentry.io/product/cli/configuration/).

mydea:
makes sense to me! Avoiding a breaking change that is easily avoidable makes sense to me. I can take care of this!

PR #109, opened 2022-11-11
feat: Use `SENTRY_HEADER` env var when available

mydea:
Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/106

Not 100% sure if the test is the "best", but couldn't think of a better way to test this easily. I'd say it is OK, but if you have another idea, I'm happy to change it.


Lms24 (approved):
LGTM 🚀

Issue #99, opened 2022-11-09
Revisit `telemetry` option and only send minimal information to Sentry


Lms24:
* [x] ~https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/111~ 
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/112
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/113
* [x] #118 
* [x] https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/120

ref #50

PR #110, opened 2022-11-11
chore(core): Remove unused dependencies

Lms24:
This PR just removes a few dependencies we don't need anymore after switching back to Sentry CLI. 

I made a small adjustment to our `captureMinimalError` function which treated Axios errors differently from other errors (we could also see this as a first step for #99)

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/telemetry.ts
+++ b/packages/bundler-plugin-core/src/sentry/telemetry.ts
@@ -7,7 +7,6 @@ import {
   NodeClient,
 } from "@sentry/node";
 import { Span } from "@sentry/tracing";
-import { AxiosError } from "axios";
 import { BuildContext } from "../types";
 
 export function makeSentryClient(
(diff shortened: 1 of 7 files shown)


mydea (approved):
This seems fine to me! Out of interest, do you have an example of what the actual error produced would be before/after this change?

Lms24:
Actually, I think I have to make a change here because we could right now capture `undefinde`. Also, I'll add a few tests to for different kinds of errors

packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -57,16 +56,22 @@ export function addSpanToTransaction(
...
-      ? {
-          name: `${isAxiosError && error.status ? error.status : ""}: ${error.name}`,
-          message: error.message,
-          stack: error.stack,
-        }
-      : {};
+export function captureMinimalError(error: unknown | Error, hub: Hub) {
+  let sentryError;
+
+  if (typeof error === "object") {
```
mydea, line 62: Ah, probably should be `error && typeof error === 'object'`, as `typeof null === 'object` (😭 )

packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -57,16 +56,22 @@ export function addSpanToTransaction(
...
+    const e = error as { name?: string; message?: string; stack?: string };
+    sentryError = {
+      name: e.name,
+      message: e.message,
+      stack: e.stack,
+    };
+  } else if (typeof error === "string") {
```
mydea, line 69: Just a though, but what if we just make this a general fallback and handle it as:

```ts
} else {
  sentryError = {
    name: 'Error',
    message: `${error}` // <-- convert anything to string
  };
}
```
}

Lms24: Sounds good to me 👍

packages/bundler-plugin-core/src/sentry/telemetry.ts
Lms24: Thanks, definitely would have missed that 😅

PR #111, opened 2022-11-11
ref(core): Don't send `org` and `project` to Sentry

Lms24:
⚠️ Needs Discussion!

We currently send the org and project as tags in each error or transaction to Sentry (if `telemetry` is enabled). Furthermore, we also set the user to `org`. 

This PR would remove all mentions or org and project in Sentry events. Happy to merge it if we say that this is necessary.

ref: #99

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -115,12 +115,8 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
   }
 
   sentryHub.setTags({
-    organization: internalOptions.org,
-    project: internalOptions.project,
     bundler: unpluginMetaContext.framework,
   });
-
-  sentryHub.setUser({ id: internalOptions.org });
 
   // This is `nonEntrypointSet` instead of `entrypointSet` because this set is filled in the `resolveId` hook and there
   // we don't have guaranteed access to *absolute* paths of files if they're entrypoints. For non-entrypoints we're


vladanpaunovic:
I find this information valuable as we could literally reach out to affected people and help them resolve the issue. On top of that, this could help us group issues and discover patterns on the user level, which in this context here is the organisation.

In my opinion, it is safe to assume that there is a consent as this is under the user opted in flag `telemetry` and is being delivered to us only for those users who opt out for using it.

mydea:
> I find this information valuable as we could literally reach out to affected people and help them resolve the issue. On top of that, this could help us group issues and discover patterns on the user level, which in this context here is the organisation.
> 
> In my opinion, it is safe to assume that there is a consent as this is under the user opted in flag `telemetry` and is being delivered to us only for those users who opt out for using it.

AFAIK this is currently not opt-in, but opt-out, meaning by default it will be on for all users.
Just thinking, what about making the `telemetry` option an enum instead of a boolean? Then we could do e.g.:

```ts
// all: Send full telemetry data, including org name
// minimal: default option, send basic telemetry data, excluding org name.
// off: No telemetry data sent
type TelemetryOption = 'all' | 'minimal' | 'off';
```

Lms24:
I'm waiting on a for a reply in `#discuss-legal` on this. If we have to make changes, the enum type would probabyl be a good solution, with the `minimal` option as a default

Lms24:
Closing in favour of #120

PR #112, opened 2022-11-11
ref(core): Don't create spans for skipped release steps

Lms24:
In our release creation pipeline, certain steps (`cleanArtifacts`, `setCommits`, `finalize`, `deploy`) can be skipped depending on the chosen options. In case they are skipped, as of this PR, we don't anymore create spans for these steps. 

This PR also adjusts release pipeline tests (and adds the steps that were previously missing)

ref #99


mydea (approved):
:shipit:

PR #113, opened 2022-11-11
feat(core): Add plugin options as tags

Lms24:
This PR adds a subset of our plugin options as tags to Sentry events. Obviously, we cannot send all options but we have to be careful to only send insensitive data, such as if an optional step (e.g. `deploy`) was selected or not. 

Going forward, we can use this data to evaluate which options are used and for debugging issues. 
As suggested by @AbhiPrasad we're sending this data as tags rather than context.

This PR also adds some tests (thanks GH copilot for the help)

ref #99

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/telemetry.ts
+++ b/packages/bundler-plugin-core/src/sentry/telemetry.ts
@@ -7,6 +7,7 @@ import {
   NodeClient,
 } from "@sentry/node";
 import { Span } from "@sentry/tracing";
+import { InternalOptions } from "../options-mapping";
 import { BuildContext } from "../types";
 
 export function makeSentryClient(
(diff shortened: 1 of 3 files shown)


packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -76,3 +77,32 @@ export function captureMinimalError(error: unknown | Error, hub: Hub) {
...
+    "add-deploy": deploy ? true : false,
+    "inject-releases-map": injectReleasesMap,
+    "dry-run": dryRun,
+    "error-handler": errorHandler ? "custom" : "none",
+  });
+
+  if (setCommits) {
+    hub.setTag("set-commits", setCommits.auto === true ? "auto" : "manual");
+  } else {
+    hub.setTag("set-commits", false);
```
AbhiPrasad, line 106: m: We can just conditionally set this tag, there’s a `has:X` query we can use for an existence check when querying in Sentry.

AbhiPrasad (approved):

mydea (approved):

packages/bundler-plugin-core/src/sentry/telemetry.ts
Lms24: Sounds good to me! [92b6638](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/113/commits/92b66383c1948754c1760e89ba2af89fd1365708)

Issue #114, opened 2022-11-14
Update READMEs to include examples


lforst:
We need examples in the READMEs of the individual bundlers/packages.

Ref: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/50

PR #115, opened 2022-11-14
docs: Add examples to individual READMEs

lforst:
Resolves https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/114

Adds examples to the READMEs of the individual bundlers.


Lms24 (approved):
m: In all "Usage" sections, we should probably add, in which file these modifications should happen, right? For Vite e.g. in `vite.config.js`, etc. 

Otherwise 🚀

PR #123, opened 2022-11-16
feat: Replace `entries` option with `releaseInjectionTargets`

lforst:
We faced problems when dogfooding the bundler plugin in https://github.com/getsentry/gib-potato.

1. We tried to inject the release in non-JS files.
    - This PR addresses this by only injecting the release code when we encounter a file that ends in a JS-like file ending (`.js`, `.ts`, `.jsx`, `.tsx`, `.cjs` or `.mjs`).
	
2. Considering the above fixed, if an app's entry point was a non-JS file (e.g. a `.html` file) we didn't inject the release anywhere at all.
    - We address this by injecting the release code in *all* user code modules (the bundlers will take care of only actually including the code once per bundle) and replacing the `entries` option with a new `releaseInjectionTargets` option to allow for more fine-grained control in overriding in which modules the release code should be injected. This is a breaking change so we added an entry in the migration docs.


MIGRATION.md
```diff
@@ -23,6 +23,15 @@ sentryWebpackPlugin({
 });
 ```
 
+### New behabiour of `entries` option
```
Lms24, line 26:
```suggestion
### New behaviour of `entries` option
```

Lms24 (approved):
🚢 it

Lms24 (approved):

PR #128, opened 2022-11-23
fix(core): Fix telemetry option logic

### LINKED ISSUE #129: Fix `telemetry` option
The telemetry option had some issues depending on what kind of format the users provided the URL in. Additionally, the logic for setting the telemetry option is kinda sketchy (manually overriding client options) so we need to refactor that as well.

lforst:
Fixes #129

- Introduces a custom transport that only sends telemetry events to our Sentry instance when Sentry CLI has determined that the target URL for the source maps is the Sentry SaaS URL. (this is due to PII reasons)
- Refactors usage of the telemetry option so we don't mutate it.
- Gets rid of manually mutating client options
- Reduces noise of log messages and breadcrumbs sent to our Sentry instance by not logging debug log statements - these are for our users to debug

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/telemetry.ts
+++ b/packages/bundler-plugin-core/src/sentry/telemetry.ts
@@ -138,11 +138,15 @@ export function addPluginOptionInformationToHub(
 }
 
 export async function shouldSendTelemetry(options: InternalOptions): Promise<boolean> {
-  const { silent, org, project, authToken, url, vcsRemote, customHeader, dist, telemetry } =
+  const { silent, org, project, authToken, url, vcsRemote, customHeader, dist, telemetry, dryRun } =
     options;
 
   // `options.telemetry` defaults to true
   if (telemetry === false) {
+    return false;
+  }
+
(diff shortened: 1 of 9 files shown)


packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -119,29 +127,61 @@ export function addPluginOptionTags(options: InternalOptions, hub: Hub) {
...
+export async function shouldSendTelemetry(options: InternalOptions): Promise<boolean> {
+  const { silent, org, project, authToken, url, vcsRemote, customHeader, dist, telemetry } =
+    options;
 
-  const url = cliInfo
-    ?.split(/(\r\n|\n|\r)/)[0]
+  // `options.telemetry` defaults to true
+  if (telemetry === false) {
+    return false;
+  }
```
Lms24, line 147: As discussed, let's also check that `dryRun` isn't isn't enabled (or return `false` if it is)

Lms24 (approved):

PR #132, opened 2022-11-23
fix(core): Normalize `id` and `releaseInjectionTargets` in `transformInclude`

lforst:
Fixes release injection for Vite on Windows.


Lms24 (approved):
Good catch!

PR #133, opened 2022-11-23
ci(e2e): Add test matrix for OS and Node.js versions

lforst:


lforst:
Implementing conditional snapshot tests is too painful at the moment. Manual testing shows it works on windows and older node versions.

PR #136, opened 2022-11-25
fix(core): Finish spans when CLI commands fail

lforst:
This PR fixes telemetry by finishing spans after CLI commands fail.

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
+++ b/packages/bundler-plugin-core/src/sentry/releasePipeline.ts
@@ -18,10 +18,13 @@ export async function createNewRelease(
 ): Promise<void> {
   const span = addSpanToTransaction(ctx, "function.plugin.create_release");
 
-  await ctx.cli.releases.new(releaseName);
+  try {
+    await ctx.cli.releases.new(releaseName);
+  } finally {
+    span?.finish();
+  }
 
   ctx.logger.info("Successfully created release.");
-  span?.finish();
 }
 
 export async function cleanArtifacts(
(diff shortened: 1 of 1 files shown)


packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -18,10 +18,13 @@ export async function createNewRelease(
 ): Promise<void> {
   const span = addSpanToTransaction(ctx, "function.plugin.create_release");
 
-  await ctx.cli.releases.new(releaseName);
+  try {
+    await ctx.cli.releases.new(releaseName);
+  } finally {
+    span?.finish();
```
AbhiPrasad, line 24: we really need a top level span method...

AbhiPrasad (approved):

packages/bundler-plugin-core/src/sentry/releasePipeline.ts
```diff
@@ -36,10 +39,13 @@ export async function cleanArtifacts(
 
   const span = addSpanToTransaction(ctx, "function.plugin.clean_artifacts");
 
-  await ctx.cli.releases.execute(["releases", "files", releaseName, "delete", "--all"], true);
+  try {
+    await ctx.cli.releases.execute(["releases", "files", releaseName, "delete", "--all"], true);
+  } finally {
+    span?.finish();
+  }
```
Lms24, line 46: Don't we put `span` onto the scope in `addSpanToTransaction`? In case we do, can we finish the span in the try/catch/finally that wraps the entire release creation pipeline?

lforst: I had that before but it felt very blanket-approachy. I like this more because it is more explicit (though admittedly repetitive).

Lms24: Fair enough, was just a thought 😅

PR #146, opened 2022-12-05
Update magic-string due to deprecated dependency

jperelli:
npm i throws this warning
```
...
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
...
```
npm ls sourcemap-codec says
```
└─┬ @sentry/esbuild-plugin@0.2.3
  └─┬ @sentry/bundler-plugin-core@0.2.3
    └─┬ magic-string@0.26.2
      └── sourcemap-codec@1.4.8
```
magic-string module [CHANGELOG.md](https://github.com/Rich-Harris/magic-string/blob/master/CHANGELOG.md) does not look like is making a breaking change from 0.26.2 to 0.27.0

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1414,7 +1414,7 @@
     "@jridgewell/gen-mapping" "^0.3.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
-"@jridgewell/sourcemap-codec@^1.4.10":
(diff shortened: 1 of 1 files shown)


lforst:
Thanks for the PR! I am gonna do some manual tests here to verify this isn't actually breaking something. Just as a heads-up, minor version bumps on `0.x` are actually considered breaking semver-spec wise.

lforst:
Looks all good! Thank you for your contribution!

PR #148, opened 2022-12-06
ref(core): Send project as `dist` in telemetry

lforst:
Sends the user project as `dist` in telemetry so we can query for it efficiently in dashboards since `dist` is an indexed parameter.

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/telemetry.ts
+++ b/packages/bundler-plugin-core/src/sentry/telemetry.ts
@@ -7,13 +7,19 @@ const SENTRY_SAAS_HOSTNAME = "sentry.io";
 
 export function makeSentryClient(
   dsn: string,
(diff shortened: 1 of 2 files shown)


packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -7,14 +7,20 @@ const SENTRY_SAAS_HOSTNAME = "sentry.io";
...
   const client = new NodeClient({
     dsn,
 
     tracesSampleRate: 1,
     sampleRate: 1,
 
+    // We're also sending the user project in dist because it is an indexed fieldso we can use this data effectively in
+    // a dashboard.
+    // Yes, this is slightly abusing the purpose of this field.
+    dist: userProject,
```
Lms24, line 22: Fine with me

Lms24 (approved):

PR #152, opened 2022-12-07
build: Add CodeQL scanning

AbhiPrasad:
This is a security requirement iirc - adding this. See: https://www.notion.so/sentry/Security-Transparency-Page-7e9937fcad3b41bd985c352ee2e10664#d584389ccb9d463b9c001f0c40915f70

PR #155, opened 2022-12-12
ci: Fix faulty image type for CodeQL

lforst:
Attempt to fix CI failing/being queued forever for CodeQL.

Ref: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/152#issuecomment-1346449182


AbhiPrasad (approved):
woops

lforst:
happens :D

PR #156, opened 2022-12-12
meta: Add chanelog entry for 0.3.0

lforst:


Lms24 (approved):
Did we drop `customHeader`? If yes, we should add an entry to the migration guide I guess

PR #164, opened 2023-01-19
deps(core): Bump unplugin version

lforst:
This PR bumps unplugin to the newest version. Even though it is a major it shouldn't have any breaking changes it is just to mark it as stable.


Lms24 (approved):
Nice!

Issue #163, opened 2023-01-18
esbuild-plugin: Error: Cannot find module 'sentry-release-injector'


Lilja:
### Environment

`@sentry/esbuild-plugin: 0.3.0`

### Steps to Reproduce
```javascript
// esbuild.config.js
const sentryEsbuildPlugin = require("@sentry/esbuild-plugin");
const buildDate = Date.now();
const cp = require("child_process");
const revision = cp.execSync("git rev-parse --short HEAD").toString().trim();
const glob = require("glob");

console.log("Sentry release", revision);

require("esbuild").build({
  entryPoints: glob.sync("./src/**/*.ts"),
  outdir: "./dist",
  bundle: false,
  sourcemap: true, // Source map generation must be turned on
  platform: "node",
  format: "cjs",
  define: {
    "process.env.SENTRY_RELEASE_ID": JSON.stringify(revision),
    "process.env.BUILT_DATE": JSON.stringify(buildDate),
    "process.env.GIT_SHA": JSON.stringify(revision),
  },
  plugins: [
    // Put the Sentry esbuild plugin after all other plugins
    sentryEsbuildPlugin.default({
      org: "funnelio",
      project: "customs-error-service",

      // Specify the directory containing build artifacts
      include: ["dist", "src"],

      // Auth tokens can be obtained from https://sentry.io/settings/account/api/auth-tokens/
      // and need `project:releases` and `org:read` scopes
      authToken: process.env.SENTRY_AUTH_TOKEN,
      release: revision,
    }),
  ],
});

`node esbuild.config.js && node ./dist/index.js`
```

### Expected Result
Valid require import

### Actual Result

```javascript
// ./dist/types.js
"use strict";
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __export = (target, all) => {
  for (var name in all)
    __defProp(target, name, { get: all[name], enumerable: true });
};
var __copyProps = (to, from, except, desc) => {
  if (from && typeof from === "object" || typeof from === "function") {
    for (let key of __getOwnPropNames(from))
      if (!__hasOwnProp.call(to, key) && key !== except)
        __defProp(to, key, { get: () => from[key], enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable });
  }
  return to;
};
[...]

PR #166, opened 2023-01-19
ref(core): Only inject release into entrypoints per default

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/163

Turns out our approach for injecting the release value sucks in case you want to preserve modules while transpiling because we'll create a throwing `require("\0sentry-release-injector")` statement since `\0sentry-release-injector` is only virtual and you can't "preserve" a virtual file.

This PR is fixing this by slightly changing how the `releaseInjectionTargets` option works. By default we'll inject the release into all entry points now (as opposed to all modules) - without any import magic - just by injecting code. This may have bundle size impact but users can still control how this works by providing a function to `releaseInjectionTargets`.


packages/bundler-plugin-core/src/index.ts
```diff
@@ -253,15 +219,22 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
...
      * @returns transformed code + source map
      */
-    transform(code, id) {
+    async transform(code, id) {
       logger.debug('Called "transform":', { id });
 
       // The MagicString library allows us to generate sourcemaps for the changes we make to the user code.
       const ms = new MagicString(code);
 
       // Appending instead of prepending has less probability of mucking with user's source maps.
```
Lms24, line 228: Is this comment still valid/necessary?

Lms24 (commented):
I like this approach. Seems to be a lot simpler than what we had before. I guess this is also kinda "behaviorally" breaking, so let's maybe add a migration note as well? 
lmk when this is ready to ✅

packages/bundler-plugin-core/src/index.ts
Lms24: (referring to line 228)

lforst: Yup thanks! :)

lforst:
@Lms24 should be ready for review!

Lms24 (approved):
Nice, let's go!

pdmshrestha:
After upgrading from 0.3.0 to 0.4.0, it seems that release is not injected property. 
```
"@sentry/tracing": "^7.38.0",
"@sentry/vite-plugin": "^0.4.0",
"@sentry/vue": "^7.38.0",
```
Didn't have any issue on v0.3.0.

I am using vite to bundle the assets. 
```
import { sentryVitePlugin } from "@sentry/vite-plugin";
```

Issue #157, opened 2022-12-12
Deprecate/delete `customHeader` in favor of `headers`


lforst:
Supporting both `headers` and `customHeaders` options is unnecessary so we should just have `headers`.

Once implemented, we shouldn't forget to add this removal/deprecation to the migration guide.

PR #167, opened 2023-01-19
ref: Remove `customHeader` option

lforst:
This PR removes the `customHeader` option in favor of `headers`.

Resolves https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/157


Lms24 (approved):
🚀

PR #171, opened 2023-01-20
feat: Add `sentryCliBinaryExists` function

lforst:
This PR adds a function which exists in the current implementation of the webpack plugin to check for the existence of the sentry cli in node modules.

We need this for the Next.js SDK.


Lms24 (approved):
LGTM!

Issue #175, opened 2023-02-15
Include names of dependencies into build


mydea:
We want to be able to capture which dependencies a project has, in order to be able to know e.g. if users are using the correct sentry package etc.

## Dependencies

Get a list of all used `dependencies` and `devDependencies` in the format:

```
['depName', 'depName2', 'depName3']
```

The list should be sorted alphabetically to ensure it is stable.

For a specified list of framework packages, we want to also include the major version. We want to consider the following packages:

* `react`
* `@angular/core`
* `vue`
* `ember-source`
* `svelte`
* `@sveltejs/kit`
* `webpack`
* `vite`
* `gatsby`
* `next`
* `remix`

This list is subject to change and may be extended later.

We want to include the major versions for these, if encountered in the app, in the format:

```
{ react: 1, next: 13 }
```

Versions should be numbers, if a major cannot be parsed, we want to omit it.

So a full dataset would be:

```
{
  deps: ['react', 'zPackage']
  depVersions: { react: 18 }
}
```

## Node version

In addition, we can also capture which node version is used, in order to have better data about what to support. The major version should be captured:

```
{
  nodeVersion: 10
}
```

## Non-goals

* We do not need to differentiate between `dependencies` and `devDependencies`
* We do not care about `peerDependencies`
* We do not care about versions for _all_ packages
* We only care about major versions for the main framework dependencies

PR #176, opened 2023-02-16
feat: Add `_experiments.injectBuildInformation` option

mydea:
This allows to gather some build information into `window.SENTRY_BUILD_INFO`.

If enabled (currently opt-in, we may flip this in a future step?), adds the following:

```js
window.SENTRY_BUILD_INFO = {
  deps: [
    "@sentry-internal/eslint-config",
    "@sentry-internal/sentry-bundler-plugin-tsconfig",
    "@sentry/bundler-plugin-core",
    "@swc/jest",
    "@types/jest",
    "@types/webpack4",
    "esbuild",
    "eslint",
    "jest",
    "rollup",
    "ts-node",
    "vite",
    "webpack",
    "webpack4",
  ],
  depsVersions: { vite: 3, webpack: 5 },
  nodeVersion: 14,
};
```

Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/175


packages/bundler-plugin-core/src/utils.ts
```diff
@@ -7,3 +12,154 @@
...
+  "@angular/core",
+  "vue",
+  "ember-source",
+  "svelte",
+  "@sveltejs/kit",
+  "webpack",
+  "vite",
+  "gatsby",
+  "next",
+  "remix",
```
AbhiPrasad, line 103: should we also do rollup/esbuild?

mydea: good point! Yes, I'll also add these

packages/bundler-plugin-core/src/index.ts
```diff
@@ -363,9 +367,30 @@ function generateGlobalInjectorCode({
       _global.SENTRY_RELEASES["${key}"]={id:"${release}"};`;
   }
 
+  if (injectBuildInformation) {
+    const buildInfo = getBuildInformation();
+
+    code += `
+      _global.SENTRY_BUILD_INFO=${JSON.stringify(buildInfo)}`;
```
lforst, line 374: h: Can we add a semicolon at the end here? Although rare, there are some situations where a missing semicolon will mess up user code.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -198,6 +198,17 @@ export type Options = Omit<IncludeEntry, "paths"> & {
...
+  /**
+   * If set to true, the plugin will inject an additional `SENTRY_BUILD_INFO` variable.
+   * This contains information about the build, e.g. dependencies, node version and other useful data.
+   *
+   * This is still experimental and subject to change!
+   *
+   * Defaults to `false` for now.
+   * @hidden
+   */
+  injectBuildInformation?: boolean;
```
lforst, line 211: l: Just a thought. Should we make this an experiment for now? Just so we don't "break" if we decide to unship? Iirc we are still not a 100% sure if we want to go this route.

lforst (approved):

packages/bundler-plugin-core/src/index.ts
mydea: oops, definitely!

packages/bundler-plugin-core/src/types.ts
mydea: Yes, probably makes sense! I'll just make it `_experiments.injectBuildInformation` for now 👍

mydea:
Updated this to be confiurable as `_experiments.injectBuildInformation` for now!

lforst (approved):

Lms24 (approved):

Issue #182, opened 2023-03-08
Automatic Release Numbering / Error correlation doesn't appear to work 


quickstark:
### Environment

"@sentry/react": "^7.38.0",
"@sentry/tracing": "^7.38.0",
"@sentry/vite-plugin": "^0.4.0",

### Steps to Reproduce

1. Don't specify a release param in vite.confit or Sentry.init
2. Build
3. Run build preview
4. Generate an Error
5. Check Sentry Issue for release value / associated sourcemap

### Expected Result

In the absence of any other release information, the Git commit hash gets automatically set as the release and correlated to any Errors within the release

### Actual Result

Git commit hash gets used, a Sentry release is created, sourcemaps are uploaded, but the Error doesn't contain the corresponding release tag.

It works as expected if you specify the release param as part of vite.config and Sentry.init.

![CleanShot 2023-03-08 at 10 48 35](https://user-images.githubusercontent.com/17515078/223843975-604b0954-b596-43ae-8fef-3ac44a7f5b75.png)

lforst:
hi what framework are you using?

smeubank:
@HazAT moving this to backlog

vite is by far our most popular plugin here, should prio it

Lms24:
Hi @quickstark thanks for reporting this! 

Can you share your `Sentry.init` config as well as your vite config with the options you pass to our Vite plugin? Also, if you can, please provide a minimal reproduction (e.g. stackblitz or a git repo) so that we can debug this. 

Thanks a lot :)

hakubo:
From what I'm able to notice is that SENTRY_RELEASE is not being injected.

Downgrading to 0.3 does help.

hakubo:
passing
```
injectReleasesMap: true
```

to configuration seems to help

hakubo:
Looks like there was a change, but not documented well: https://github.com/getsentry/sentry-javascript-bundler-plugins/commit/6ca81b8b0657bb210623e0dab8f8eefaec7e5eb6

hakubo:
Documentation stills states that: 'By default, the release will be injected into all modules - however, bundlers will include the injected release code only once per entrypoint. If release injection should be disabled, provide an empty array here.'

hakubo:
I susspect the issue is here: 

resolveId(id, importer, { isEntry }) {

[...]

Issue #180, opened 2023-03-01
Release injection not working with @sentry/plugin-vite


dd-jonas:
### Environment

- `vite @4.1.4`
- `@sentry/vite-plugin @0.4.0`

### Steps to Reproduce

1. Add Sentry plugin in `vite.config.ts`
1. Run `vite build` to trigger a Sentry release
1. Run `vite preview`
1. Log `SENTRY_RELEASE` in the Chrome devtools console

### Expected Result

`{ id: "abc123" }`

### Actual Result

`Uncaught ReferenceError: SENTRY_RELEASE is not defined`

### Additional info

I did some debugging and it looks like the index.html file is seen as the entry point of the application. This means that the `transformInclude` function will not include any file as the .html extension is not valid, which means the transform hook won't get called at all and no release id is injected. Not sure what the reason is why the index.html is the entry point. Currently working around this by using `releaseInjectionTargets: /src\/index\.tsx$/`.

andrey-hohlov:
https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/178#issuecomment-1440741634

dd-jonas:
Cool, good to know that `releaseInjectionTargets` is the recommended way of handling this

lforst:
I thought of a way to fix release injection for this use case but I won't get to it for a while. Give or take a month.

In the meanwhile `releaseInjectionTargets` is definitely the way to go.

PR #188, opened 2023-03-28
feat(core): Import release injection code from each module

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/182
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/180

This somewhat reverts to an old behavior where we inject the release injector code into each module. The difference this time is that we reference a file that actually exists in our package instead of a virtual file. This should circumvent the situation we ran into in https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/163 because the release injector module will now always exist.


packages/bundler-plugin-core/package.json
```diff
@@ -10,14 +10,9 @@
...
-    "dist"
+    "dist",
+    "sentry-release-injection-file.js"
   ],
-  "exports": {
-    ".": {
-      "import": "./dist/esm/index.mjs",
-      "require": "./dist/cjs/index.js"
-    }
-  },
```
lforst, line 20: This was causing weird problems when importing `@sentry/bundler-plugin-core/sentry-release-injection-file`. This standard is cursed.

AbhiPrasad: It might be that with this you have to name it `sentry-release-injection-file.mjs` for esm.

AbhiPrasad (approved):

Issue #186, opened 2023-03-24
Allow release injection or upload only tasks


alexandresoro:
Hello,

I’d like to suggest an improvement to be able the configure the plugins to be able to do either release injection but no source upload or the opposite.

It may already be possible to do that by tweaking some options like `include` but I am unsure of the side effects.

As far as I understand, this could be in line with the rationale described in the comments here : https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/97a2a2f7e72b19aade44d008a55ee5a8490c51b1/packages/bundler-plugin-core/src/index.ts#L32

The use case that would be improved is the following:
A CI tool is used to build and upload the source maps.
Another tool e.g. edge deployment is used for deployment only.

The CI tool needs to be able to create and upload source maps. However, the edge deployment will usually only need release injection but is not expected to build or deploy the source maps.

Currently, I use the edge deployment to build and upload source maps as I need to inject the release there.
I also have a small script to remove the maps before deployment to prevent them from being deployed, but this is a workaround.
If both tasks could be done one without the other, I could move the sourcemap pipeline to the CI tool and benefit from additional features like commit tracking. And that would leave the edge deployment to the only thing it needs: release injection.
Obviously, in this use case, this is possible because of the release name resolution on a Git repo that would guarantee that both builds of the same level of code would produce the same release name.

alexandresoro:
Yes, that would be seem fitting!

Related to `disableSourceMapsUpload` I found out that I could achieve something similar by setting `include` to `""`. So this does not upload anything.
However, it still tries to connect to the Sentry destination instance.
So, as far as I can see a potential `disableSourceMapsUpload` flag should also make sure that no connection is attempted to Sentry - and no env var related to that is necessary.

lforst:
Seems like a fair suggestion. I'll put it onto our backlog. PRs are welcome!

alexandresoro:
> Seems like a fair suggestion. I'll put it onto our backlog. PRs are welcome!

Thank you! I'll try to have a look and draft something!

alexandresoro:
I've drafted a PR that implemented both, though I am unsure that the proposed PR is optimal.

PR #190, opened 2023-03-28
feat(core): Allow user to disable release injection or source maps upload

alexandresoro:
Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/186

This PR exposes two new user options:
- `disableReleaseInjection`: if set to `true`, it won't add any release information into the bundled files.
- `disableSourceMapsUpload`: if set to `true`, the process to upload generated source maps to Sentry will be skipped.

Tests have been updated accordingly.

Regarding `disableReleaseInjection`, the proposed solution is to be "transparent" related to the `transform` action. It will simply return what was passed as input and bypass the `prepend` step.
This may not be optimal, but I did not manage to find any less intrusive solution, as completely disabling the transformation seem to also disable the rest of the process, which may still be desired.


lforst (approved):

lforst:
Thank you very much for taking the time to implement this! :)

Issue #191, opened 2023-03-29
Add debug Id injection


lforst:

smeubank:
let's review what all needs to be changed for docs

PR #192, opened 2023-03-30
feat(core): Add experimental debug ID based source map upload to Rollup and Vite plugins

lforst:
Ref https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/191

This PR adds an experimental option to the bundler plugins that will upload source files with debug IDs to Sentry.

In short, the following steps are executed:
- Inject a code snippet into every chunk/bundle that will put a debug ID for that particular chunk/bundle onto a global object. Additionally, this code snippet contains a string with a certain format that will allow us to extract that debug ID in later steps.
- Create a temp folder that will be used to upload files with debug IDs to Sentry.
- Inject a `//# debugId=...` comment at the bottom of the chunks and write the chunk to the temp folder.
- Write a `debugId` and `debug_id` field to the corresponding source maps if they are found and write them to the temp folder.
- Upload the temp folder via Sentry CLI with an option that enables artifact uploads.
- Delete the temp folder.

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id.ts
+++ b/packages/bundler-plugin-core/src/debug-id.ts
@@ -78,8 +78,7 @@ export async function prepareBundleForDebugIdUpload(
     }
   });
 
-  await writeSourceFilePromise;
-  await writeSourceMapFilePromise;
+  return Promise.all([writeSourceFilePromise, writeSourceMapFilePromise]);
 }
 
 /**
--- a/packages/bundler-plugin-core/src/debug-id.ts
+++ b/packages/bundler-plugin-core/src/debug-id.ts
@@ -1,0 +1,149 @@
+import * as fs from "fs";
+import MagicString from "magic-string";
+import * as path from "path";
+import * as util from "util";
+import { Logger } from "./sentry/logger";
+import { stringToUUID } from "./utils";
+
+// TODO: Find a more elaborate process to generate this. (Maybe with type checking and built-in minification)
+const DEBUG_ID_INJECTOR_SNIPPET =
+  ';!function(){try{var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof self?self:{},n=(new Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="__SENTRY_DEBUG_ID__",e._sentryDebugIdIdentifier="sdbid-__SENTRY_DEBUG_ID__")}catch(e){}}();';
+
+export function injectDebugIdSnippetIntoChunk(code: string) {
+  const debugId = stringToUUID(code); // generate a deterministic debug ID
+  const ms = new MagicString(code);
+
+  const codeToInject = DEBUG_ID_INJECTOR_SNIPPET.replace(/__SENTRY_DEBUG_ID__/g, debugId);
+
+  // We need to be careful not to inject the snippet before any `"use strict";`s.
+  // As an additional complication `"use strict";`s may come after any number of comments.
+  const commentUseStrictRegex =
+    /^(?:\s*|\/\*(.|\r|\n)*?\*\/|\/\/.*?[\n\r])*(?:"use strict";|'use strict';)?/;
+
+  if (code.match(commentUseStrictRegex)?.[0]) {
+    // Add injected code after any comments or "use strict" at the beginning of the bundle.
+    ms.replace(commentUseStrictRegex, (match) => `${match}${codeToInject}`);
+  } else {
+    // ms.replace() doesn't work when there is an empty string match (which happens if
+    // there is neither, a comment, nor a "use strict" at the top of the chunk) so we
+    // need this special case here.
+    ms.prepend(codeToInject);
+  }
+
+  return {
+    code: ms.toString(),
+    map: ms.generateMap(),
+  };
+}
+
+export async function prepareBundleForDebugIdUpload(
+  bundleFilePath: string,
+  uploadFolder: string,
+  uniqueUploadName: string,
+  logger: Logger
+) {
+  let bundleContent;
+  try {
+    bundleContent = await util.promisify(fs.readFile)(bundleFilePath, "utf8");
+  } catch (e) {
+    logger.warn(`Could not read bundle to determine debug ID and source map: ${bundleFilePath}`);
+    return;
+  }
+
+  const debugId = determineDebugIdFromBundleSource(bundleContent);
+  if (debugId === undefined) {
+    logger.warn(`Could not determine debug ID from bundle: ${bundleFilePath}`);
+    return;
+  }
+
+  bundleContent += `\n//# debugId=${debugId}`;
+  const writeSourceFilePromise = util.promisify(fs.writeFile)(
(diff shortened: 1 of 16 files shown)


packages/bundler-plugin-core/src/debug-id.ts
```diff
@@ -0,0 +1,149 @@
+import * as fs from "fs";
+import MagicString from "magic-string";
+import * as path from "path";
+import * as util from "util";
+import { Logger } from "./sentry/logger";
+import { stringToUUID } from "./utils";
+
+// TODO: Find a more elaborate process to generate this. (Maybe with type checking and built-in minification)
+const DEBUG_ID_INJECTOR_SNIPPET =
+  ';!function(){try{var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof self?self:{},n=(new Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="__SENTRY_DEBUG_ID__",e._sentryDebugIdIdentifier="sdbid-__SENTRY_DEBUG_ID__")}catch(e){}}();';
```
AbhiPrasad, line 10: We can build this whenever we need to change it - same approach as we use here: https://github.com/getsentry/sentry/pull/46780. I don't think we need to automate this, we can just manually run terser.

So everytime this snippet changes, we rebuild the minified JS. We can have this in follow up PR.

packages/bundler-plugin-core/src/utils.ts
```diff
@@ -165,3 +166,23 @@ function lookupPackageJson(cwd: string, stopAt: string): PackageJson | undefined
   const newCwd = path.dirname(path.resolve(jsonPath + "/.."));
   return lookupPackageJson(newCwd, stopAt);
 }
+
+/**
+ * Deterministically hashes a string and turns the hash into a uuid.
+ */
+export function stringToUUID(str: string): string {
```
AbhiPrasad, line 173: l: Can we have a test that enforces this will return a uuid - sanity check mostly that it behaves ok across Node versions.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -214,17 +214,32 @@ export type Options = Omit<IncludeEntry, "paths"> & {
...
+     * Defaults to `false`.
+     */
     injectBuildInformation?: boolean;
+
+    /**
+     * Configuration for debug ID upload.
+     *
+     * Note: Currently only functional for Vite and Rollup.
+     */
+    debugIdUpload?: {
```
AbhiPrasad, line 233: m: Can we add the default values for include/ignore?

packages/bundler-plugin-core/src/debug-id.ts
```diff
@@ -0,0 +1,163 @@
...
+  return undefined;
+}
+
+/**
+ * Reads a source map, injects debug ID fields, and writes the source map to the target path.
+ */
+async function prepareSourceMapForDebugIdUpload(
+  sourceMapPath: string,
+  targetPath: string,
+  degugId: string,
```
AbhiPrasad, line 143:
```suggestion
  debugId: string,
```

packages/bundler-plugin-core/src/debug-id.ts
```diff
@@ -0,0 +1,163 @@
...
+        sourceMapPath,
+        path.join(uploadFolder, `${uniqueUploadName}.js.map`),
+        debugId,
+        logger
+      );
+    }
+  });
+
+  await writeSourceFilePromise;
+  await writeSourceMapFilePromise;
```
AbhiPrasad, line 82: l: Why double await here? If we want it to run at the same time, can we just do `Promise.allSettled`?

packages/bundler-plugin-core/src/debug-id.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * Looks for a particular string pattern (`sdbid-[debug ID]`) in the bundle
```
AbhiPrasad, line 86: Why did we chose `sdbid`?

AbhiPrasad (commented):
This needs tests, especially around the `use strict` logic and determining source map paths.

packages/bundler-plugin-core/src/debug-id.ts
lforst: Just wanted to use something short. It can be changed at any time since nothing depends on this. Do you think another pattern fits better?

AbhiPrasad: I would rather it be `sentry`, but it's no big deal.

packages/bundler-plugin-core/src/utils.ts
lforst: Added a small test: 4154b4a974e30183befed53d36674a4e80754147

packages/bundler-plugin-core/src/types.ts
lforst: I am not sure if a default for include makes sense, since you either define paths or you don't. Ignore makes sense though, so I added it: a7cf4e35dbaefc20a93aedd0c2f99063eee415bc

packages/bundler-plugin-core/src/debug-id.ts
lforst: 22f56d9092b8e8a87e95edd6fb4c7726b8ff3a0a

packages/bundler-plugin-core/src/debug-id.ts
lforst: I think we want to throw here in case one of the promises fails so I went with `Promise.all`: 025e4c1db1ae87148391cba01552d49ffa5bd6cd

packages/bundler-plugin-core/src/debug-id.ts
lforst: > I would rather it be sentry, but it's no big deal.

No! good point! :) befc007f1c4399d3d91a2d7dec1ae3c34bc3959a

packages/bundler-plugin-core/test/utils.test.ts
```diff
@@ -169,3 +169,9 @@ describe("getDependencies", () => {
     });
   });
 });
+
+describe("stringToUUID", () => {
+  test("should return a deterministic UUID", () => {
+    expect(stringToUUID("Nothing personnel kid")).toBe("52c7a762-5ddf-49a7-6f16-6874a8cb2512");
```
AbhiPrasad, line 175: ❤️

packages/bundler-plugin-core/src/debug-id.ts
```diff
@@ -0,0 +1,162 @@
...
+  try {
+    const sourceMapFileContent = await util.promisify(fs.readFile)(sourceMapPath, {
+      encoding: "utf8",
+    });
+
+    const map = JSON.parse(sourceMapFileContent) as Record<string, string>;
+
+    // For now we write both fields until we know what will become the standard - if ever.
+    map["debug_id"] = debugId;
+    map["debugId"] = debugId;
```
AbhiPrasad, line 154: image

AbhiPrasad (approved):
let's ship and try testing it out!

Issue #194, opened 2023-04-05
@sentry/vite-plugin 0.5.0 breaks react build


SteinRobert:
### Environment

@sentry-vite 0.5.0

### Steps to Reproduce

1. Bumped @sentry/vite-plugin from 0.4.0 to 0.5.0
2. Build react app
3. 
```
#22 5.795 "default" is not exported by "node_modules/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js", imported by "node_modules/@sentry/react/esm/profiler.js".
#22 5.795 file: /ui/node_modules/@sentry/react/esm/profiler.js:3:0
#22 5.795 1: import { getCurrentHub } from '@sentry/browser';
#22 5.795 2: import { timestampWithMs } from '@sentry/utils';
#22 5.795 3: import hoistNonReactStatics from 'hoist-non-react-statics';
#22 5.795    ^
#22 5.795 4: import * as React from 'react';
#22 5.795 5: import { REACT_MOUNT_OP, REACT_UPDATE_OP, REACT_RENDER_OP } from './constants.js';
#22 5.797 error during build:
#22 5.797 RollupError: "default" is not exported by "node_modules/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js", imported by "node_modules/@sentry/react/esm/profiler.js".
```

### Expected Result

Successful compilation.

### Actual Result

-

danawoodman:
Noticed the same issue in a SvelteKit project with the dependency `debug`. Spend hours trying to track it down as I couldn't imagine the Sentry plugin could have anything to do with it 🤦

micah-redwood:
Also having builds fail (`vite build`) after upgrading from 0.4.0 to 0.5.0

```
vite v4.1.4 building for qa...
[sentry-vite-plugin] Info: Sending error and performance telemetry data to Sentry.
[sentry-vite-plugin] Info: To disable telemetry, set `options.telemetry` to `false`.
✓ 3473 modules transformed.
Cannot read properties of undefined (reading 'slice')
error during build:
TypeError: Cannot read properties of undefined (reading 'slice')
    at getCodeFrame (file:///{edited}/node_modules/rollup/dist/es/shared/node-entry.js:2013:65)
    at augmentCodeLocation (file:///{edited}/node_modules/rollup/dist/es/shared/node-entry.js:2122:28)
[...]

PR #195, opened 2023-04-06
fix(core): Skip all transformations for 3rd party modules

lforst:
Hopefully fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/194

It seems like injecting into 3rd party modules may cause them to be bundled/resolved differently and this will fail some builds.

It could be that this is just an incompatibility with another plugin but I say it is worth just making this exemption.


AbhiPrasad (approved):
Were you able to reproduce?

lforst:
> Were you able to reproduce?

@AbhiPrasad  Yup will leave a comment in the original issue!

PR #198, opened 2023-04-07
feat(webpack): Add debug ID injection to the webpack plugin

lforst:
Ref: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/191

Similar to https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/192, we now add debug ID injection to the Webpack plugin.

Note this needs tests. We will add them as a follow-up when we identified some quirks with the injection across all bundlers.

### DIFF
--- a/packages/bundler-plugin-core/package.json
+++ b/packages/bundler-plugin-core/package.json
@@ -43,7 +43,7 @@
     "glob": "9.3.2",
     "magic-string": "0.27.0",
     "unplugin": "1.0.1",
-    "webpack-sources": "^3.2.3"
+    "webpack-sources": "3.2.3"
   },
   "devDependencies": {
     "@babel/core": "7.18.5",
--- a/packages/bundler-plugin-core/package.json
+++ b/packages/bundler-plugin-core/package.json
@@ -42,7 +42,8 @@
     "find-up": "5.0.0",
     "glob": "9.3.2",
     "magic-string": "0.27.0",
-    "unplugin": "1.0.1"
+    "unplugin": "1.0.1",
+    "webpack-sources": "^3.2.3"
   },
   "devDependencies": {
     "@babel/core": "7.18.5",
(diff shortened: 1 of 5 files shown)


packages/bundler-plugin-core/package.json
```diff
@@ -42,7 +42,8 @@
     "find-up": "5.0.0",
     "glob": "9.3.2",
     "magic-string": "0.27.0",
-    "unplugin": "1.0.1"
+    "unplugin": "1.0.1",
+    "webpack-sources": "^3.2.3"
```
AbhiPrasad, line 46: m: should we hard pin? Also this doesn't change `yarn.lock`?

packages/bundler-plugin-core/src/index.ts
```diff
@@ -389,6 +391,56 @@ const unplugin = createUnplugin<Options>((options, unpluginMetaContext) => {
...
+                const source = compilation.assets[fileName];
+
+                if (!source) {
+                  logger.warn(
+                    "Unable to access compilation assets. If you see this warning, it is likely a bug in the Sentry webpack plugin. Feel free to open an issue at https://github.com/getsentry/sentry-javascript-bundler-plugins with reproduction steps."
+                  );
+                  return;
+                }
+
+                compilation.updateAsset(fileName, () => {
```
AbhiPrasad, line 410: do we need to cache here? So we only inject once?

AbhiPrasad (approved):

packages/bundler-plugin-core/package.json
lforst: good point. Added. The yarn lock didn't change because unplugin itself depended on the exact same version.

packages/bundler-plugin-core/src/index.ts
lforst: Added the cache inspired by https://github.com/webpack/webpack/pull/15454

PR #200, opened 2023-04-07
fix(core): Don't exclude release injection module

hakubo:
sentry-release-injection-file is coming from node_modules so you have to include that before you exclude node_modules.


lforst:
Thank you for catching and fixing this! I guess this slipped through because our tests use yarn workspaces/linking. We probably should add some proper E2E tests like we have in https://github.com/getsentry/sentry-javascript

lforst (approved):

Issue #211, opened 2023-04-17
require is not defined in ES module scope, you can use import instead


MathiasWP:
### Environment
`"@sentry/vite-plugin": "^0.7.1"`

### Steps to Reproduce
I followed the exact steps from https://docs.sentry.io/platforms/node/guides/connect/sourcemaps/uploading/vite/

### Expected Result
Project should build

### Actual Result
```bash
failed to load config from /Users/app/vite.config.ts
error when starting dev server:
ReferenceError: require is not defined in ES module scope, you can use import instead
    at file:///Users/@sentry/bundler-plugin-core/dist/esm/index.mjs:2121:32
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:527:24)
    at async loadConfigFromBundledFile (file:///Users/frontend/node_modules/vite/dist/node/chunks/dep-79892de8.js:63967:21)
    at async loadConfigFromFile (file:///Users/frontend/node_modules/vite/dist/node/chunks/dep-79892de8.js:63852:28)
    at async resolveConfig (file:///Users/frontend/node_modules/vite/dist/node/chunks/dep-79892de8.js:63466:28)
    at async createServer (file:///Users/frontend/node_modules/vite/dist/node/chunks/dep-79892de8.js:62752:20)
    at async CAC.<anonymous> (file:///Users/frontend/node_modules/vite/dist/node/cli.js:734:24)
```

lforst:
Hey, can you check if this is an issue when downgrading the plugin to version `0.6.0`?

henrik-d:
I have the same problem with 0.7.1. Version 0.6.0 is working though.

lforst:
Alright we published another patch. Can you give it another go? This time with version `0.7.2`? :)

Also, I wanna encourage everybody to use the new `sourcemaps.assets` option. It should be much more resilient than the `include` option.

henrik-d:
Wow, that was fast. For me this fixes the issue.

lforst:
<img width="210" alt="Screenshot 2023-04-17 at 20 23 28" src="https://user-images.githubusercontent.com/8118419/232576361-c855c150-3449-420b-bf80-9e1bb3cf6d0f.png">

MathiasWP:
I updated, but now I'm met with this error:

```bash
[...]

PR #212, opened 2023-04-17
fix(core): Use `createRequire` to not use built-in `require` in ESM

lforst:
fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/211

Rollup takes care of correctly transpiling the import meta stuff


Lms24 (approved):
Damn, ESM/CJS strikes again...

PR #218, opened 2023-04-28
ref(core): Extract release injection into separate plugins

lforst:
Extracts the release injection logic into individual bundler-specific plugins.

### Why?

Release injection messes with module/bundle code and is quite finnicky to get working with all bundlers. Unplugin doesn't seem to help in a lot of cases and even introduces additional bugs with source maps.

Cool side fact: With this change we will always be injecting at the top of modules/bundles, meaning the release value will always be available. This has not been the case for the Sentry webpack plugin so far.


lforst:
This is unfortunately a bigger one :(

AbhiPrasad (approved):
in the end its always dependency injection

PR #220, opened 2023-04-28
ci: Use setup-node instead of volta

lforst:
Volta keeps flaking and rate-limiting :(

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,7 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


mydea (approved):

.github/workflows/checks.yml
```diff
@@ -42,7 +48,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: volta-cli/action@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: "package.json"
```
AbhiPrasad, line 53: I don't think we need this, setup node should use the volta version automatically.

PR #221, opened 2023-04-28
test: Remove tests that test deprecated features

lforst:
`releaseInjectionTargets` will not be an option in future versions for technical reasons (we cannot make it work in esbuild).


AbhiPrasad (approved):
tbh I'd rather we delete the tests later, but I won't unblock.

lforst:
> tbh I'd rather we delete the tests later, but I won't unblock.

I agree but I wanna make life easier for myself right now because everything is just pure pain and I don't plan on releasing anything else than the next major right now.

PR #229, opened 2023-05-02
ref(core): Use factory function to create individual plugins

lforst:
This is the first step to separate concerns between plugins (e.g. so that we don't have to import webpack for the rollup plugin)


mydea (approved):

AbhiPrasad (approved):
Looking forward to the `sentryUnpluginFactoryFactory`

PR #232, opened 2023-05-03
ref(core): Extract release management into a separate plugin

lforst:
Part of larger cleanup


AbhiPrasad (approved):
you're really racking up those line change numbers good work.

PR #237, opened 2023-05-05
ref(core): Switch to v2 options

lforst:
Switches to the new options as decided in https://github.com/getsentry/rfcs/blob/main/text/0086-sentry-bundler-plugins-api.md


AbhiPrasad (approved):
Nice!

packages/bundler-plugin-core/src/index.ts
```diff
@@ -121,53 +118,87 @@ export function sentryUnpluginFactory({
...
+    if (options.sourcemaps) {
+      if (!options.authToken) {
+        logger.warn("No auth token provided. Will not upload source maps.");
+      } else if (!options.org) {
+        logger.warn("No org provided. Will not upload source maps.");
+      } else if (!options.project) {
+        logger.warn("No project provided. Will not upload source maps.");
+      } else if (!options.sourcemaps.assets) {
+        logger.warn("No assets defined. Will not upload source maps.");
+      } else {
```
Lms24, line 178: l: Personally I'd prefer to extract this validation to a function (per plugin) to declutter this one a little but feel free to leave it as is as.

packages/bundler-plugin-core/src/index.ts
```diff
@@ -121,53 +118,87 @@ export function sentryUnpluginFactory({
...
       plugins.push(releaseInjectionPlugin(injectionCode));
     }
 
-    if (options.sourcemaps?.assets) {
-      plugins.push(debugIdInjectionPlugin());
-    }
-
-    if (releaseName) {
+    if (!options.release.name) {
+      logger.warn("No release name provided. Will not manage release.");
```
Lms24, line 134: l: Hmm not sure how we're communitcating this plugin in the docs/readme in the future, so feel free to disregard if you think this makes sense. But IMO as a user I'd be confused what this message means. Let's make it clear that we're talking about releases on Sentry.

I'm not too sure about a better wording but what about something along these lines:

```suggestion
      logger.warn("No release name provided. Will not create release on Sentry.");
```

or

```suggestion
      logger.warn("No release name provided. Will not create a Sentry release.");
```

packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -60,39 +60,30 @@ export function createSentryInstance(
...
-  if (finalize) {
+  if (release.finalize) {
     hub.setTag("finalize-release", true);
   }
-  if (deploy) {
+  if (release.deploy) {
     hub.setTag("add-deploy", true);
   }
 
   // Miscelaneous options
```
Lms24, line 86: super-l: Should we add a tag for the new deleteAfterUpload option? Might be interesting to see how often it is used?

Lms24 (approved):
Looks good to me! Just had some minor comments but nothing blocking.

packages/bundler-plugin-core/src/index.ts
Lms24: Oh I just realized, you addressed this in https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/240 - will mark this as resolved then!

packages/bundler-plugin-core/src/sentry/telemetry.ts
lforst: We don't have it implemented yet but it probably doesn't hurt to add right now!

packages/bundler-plugin-core/src/types.ts
```diff
@@ -449,27 +198,24 @@ export interface NewOptions {
      */
     deploy?: DeployOptions;
 
+    /**
+     * Remove all the artifacts in the release before the upload.
+     *
+     * Defaults to `false`.
+     */
+    cleanArtifacts?: boolean;
```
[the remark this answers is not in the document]
lforst: Good point!

packages/bundler-plugin-core/src/index.ts
lforst: This currently has a use for TS. We could put it into a helper but I am not sure if it will actually simplify the code by a lot. Gonna leave this for now.

PR #240, opened 2023-05-08
ref(core): Improve log messages

lforst:


packages/bundler-plugin-core/src/plugins/debug-id-upload.ts
```diff
@@ -207,21 +209,33 @@ async function prepareSourceMapForDebugIdUpload(
...
   logger: Logger
 ): Promise<void> {
+  let sourceMapFileContent: string;
   try {
-    const sourceMapFileContent = await util.promisify(fs.readFile)(sourceMapPath, {
+    sourceMapFileContent = await util.promisify(fs.readFile)(sourceMapPath, {
       encoding: "utf8",
     });
+  } catch (e) {
+    logger.error(`Failed to read source map for debug ID upload: ${sourceMapPath}`);
```
AbhiPrasad, line 218: should we also log out the errors here? Might help debugging.

AbhiPrasad (approved):

Lms24 (approved):

PR #243, opened 2023-05-10
feat(core): Implements rewrite sources for debug ID upload

lforst:
Adds the `rewriteSources` hook to the plugin which allows to modify the names of files inside the stacktrace on Sentry.

The hook is called for each individual source file and also passed the entire source map object for convenience.

The default logic if not otherwise defined is to make all the sources relative to `process.cwd()` or to strip away a protocol if there is one (this is mainly for webpack).


Lms24 (approved):
l (not required, just a thought): Should we add tests for this (and maybe also for the deleteSourceMaps option)?

lforst:
> l (not required, just a thought): Should we add tests for this (and maybe also for the deleteSourceMaps option)?

Definitely but I am gonna do that later

PR #244, opened 2023-05-14
feat(core): Add `deleteFilesAfterUpload` option

### LINKED ISSUE #147: Option to automatically delete source maps after upload
For security purposes, we should not expose sourcemaps online. So a `cleanSourceMaps` configuration is good to autodelete sourcemaps after artifacts has been uploaded successfully.

lforst:
Fixes #147

### DIFF
--- a/packages/bundler-plugin-core/src/types.ts
+++ b/packages/bundler-plugin-core/src/types.ts
@@ -125,10 +125,8 @@ export interface Options {
      * The globbing patterns follow the implementation of the `glob` package. (https://www.npmjs.com/package/glob)
      *
      * Use the `debug` option to print information about which files end up being deleted.
-     *
-     * @hidden Not yet implemented.
      */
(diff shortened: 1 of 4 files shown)


packages/bundler-plugin-core/src/types.ts
```diff
@@ -125,10 +125,8 @@ export interface Options {
      * The globbing patterns follow the implementation of the `glob` package. (https://www.npmjs.com/package/glob)
      *
      * Use the `debug` option to print information about which files end up being deleted.
-     *
-     * @hidden Not yet implemented.
      */
-    deleteAfterUpload?: string | string[];
+    deleteFilesAfterUpload?: string | string[];
```
Lms24, line 129: The default is `undefined` at the moment, right? wdyt about defaulting to the source maps assets glob path(s) instead?

Lms24 (approved):
Nice!

PR #246, opened 2023-05-15
ref(core): Make better use of Sentry

lforst:
With this PR we will use sessions in the telemetry to measure how many runs failled.


packages/bundler-plugin-core/src/plugins/telemetry.ts
```diff
@@ -1,35 +1,30 @@
...
-    },
-    async writeBundle() {
-      pluginExecutionTransaction.finish();
-      await sentryClient.flush();
+    async buildStart() {
+      if (await shouldSendTelemetry) {
+        logger.info(
+          "Sending error and performance telemetry data to Sentry. To disable telemetry, set `options.telemetry` to `false`."
+        );
+        sentryHub.startTransaction({ name: "Sentry Bundler Plugin execution" }).finish();
```
Lms24, line 25: m: What is this for?

lforst: Sessions can't hold tags. So we need to send transactions. This is the same transaction we had before, but without any timing data. With the new architecture (and actually in general) timing bundler runs is unreliable und useless data. If we want to track specifics (like how long the upload takes), we should create separate transactions for that.

packages/bundler-plugin-core/src/index.ts
```diff
@@ -86,16 +92,25 @@ export function sentryUnpluginFactory({
...
+          }
         } else {
-          options.errorHandler(new Error("An unknown error occured"));
+          sentrySession.status = "crashed";
+          throw unknownError;
         }
-      } else {
-        throw unknownError;
+      } finally {
+        sentryHub.endSession();
```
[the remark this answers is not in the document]
lforst: > Do we need this in addition to the process.on endSession call?

Yes. beforeExit is not called on exceptions and the like: https://nodejs.org/api/process.html#event-beforeexit

> Do we need this in addition to the process.on endSession call?

No because that variable is only there to prevent an infinite loop with beforeExit which will be invoked again when the event loop is empty.

I'll add comments! :)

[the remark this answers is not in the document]
Lms24: thx for explaining :)

packages/bundler-plugin-core/src/plugins/telemetry.ts
Lms24: I see, makes sense, thx

Lms24 (approved):

PR #251, opened 2023-05-19
fix(core): Account for undefined release name values

lforst:
Fixes a bug where Sentry CLI throws if we only pass an empty string as a release value for debug ID upload.

### DIFF
--- a/packages/bundler-plugin-core/src/plugins/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/plugins/debug-id-upload.ts
@@ -87,16 +87,19 @@ export function debugIdUploadPlugin({
           })
         );
 
-        await cliInstance.releases.uploadSourceMaps(releaseName ?? "", {
-          include: [
-            {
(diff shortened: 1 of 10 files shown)


packages/bundler-plugin-core/src/plugins/debug-id-upload.ts
```diff
@@ -87,16 +87,19 @@ export function debugIdUploadPlugin({
...
-            {
-              paths: [tmpUploadFolder],
-              rewrite: false,
-              dist: dist,
-            },
-          ],
-          useArtifactBundle: true,
-        });
+        await cliInstance.releases.uploadSourceMaps(
+          releaseName ?? "undefined", // unfortunetly this needs a value for now but it will not matter since debug IDs overpower releases anyhow
```
Lms24, line 91: Are we/CLI team going to fix this on the CLI side ?

Lms24 (approved):

PR #257, opened 2023-05-23
feat: Auto detect build artifacts

lforst:
Will auto detect build artifacts based on the information bundlers provide in the build end hooks - making the `assets` option optional.

Risky change incoming.

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -64,7 +64,7 @@ export function createDebugIdUploadFunction({
         globAssets = assets;
       } else {
         logger.debug(
-          "Not `sourcemaps.assets` option provided, falling back to uploading detected build artifacts."
+          "No `sourcemaps.assets` option provided, falling back to uploading detected build artifacts."
         );
         globAssets = buildArtifactPaths;
       }
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -59,8 +59,18 @@ export function createDebugIdUploadFunction({
 
       folderToCleanUp = tmpUploadFolder;
 
+      let globAssets;
+      if (assets) {
+        globAssets = assets;
+      } else {
+        logger.debug(
+          "Not `sourcemaps.assets` option provided, falling back to uploading detected build artifacts."
+        );
+        globAssets = buildArtifactPaths;
+      }
+
       const debugIdChunkFilePaths = (
-        await glob(assets ?? buildArtifactPaths, {
+        await glob(globAssets, {
           absolute: true,
           nodir: true,
           ignore: ignore,
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -3,8 +3,7 @@ import { glob } from "glob";
 import os from "os";
 import path from "path";
 import * as util from "util";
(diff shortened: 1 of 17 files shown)


packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -47,34 +46,41 @@ export function debugIdUploadPlugin({
...
+          ignore: ignore,
+        })
+      ).filter(
+        (debugIdChunkFilePath) =>
+          debugIdChunkFilePath.endsWith(".js") ||
+          debugIdChunkFilePath.endsWith(".mjs") ||
+          debugIdChunkFilePath.endsWith(".cjs")
+      );
 
+      if (Array.isArray(assets) && assets.length === 0) {
```
Lms24, line 75: Is `assets` always normalized to an array? otherwise, what about `assets: ""`?

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -47,34 +46,41 @@ export function debugIdUploadPlugin({
...
-            ignore: ignore,
-          })
-        ).filter(
-          (debugIdChunkFilePath) =>
-            debugIdChunkFilePath.endsWith(".js") ||
-            debugIdChunkFilePath.endsWith(".mjs") ||
-            debugIdChunkFilePath.endsWith(".cjs")
-        );
+      const debugIdChunkFilePaths = (
+        await glob(assets ?? buildArtifactPaths, {
```
Lms24, line 63: Can we add a log that we're using the auto-determined paths if we fall back to `buildArtifactPaths`?

Lms24 (commented):
I agree it's a risky change but the implementation makes sense to me. Nevertheless, I think we should cover this with at least one test, wdyt?

packages/bundler-plugin-core/src/debug-id-upload.ts
lforst: debug or info level?

packages/bundler-plugin-core/src/debug-id-upload.ts
lforst: My thought process here was to just skip uploading entirely when users provide an empty array (as a means of disabling I guess). An empty string is more or less user error which we don't care to catch here. I think this will be implicitly caught by the glob pattern not finding anything and the message below then saying "check your assets option".

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -47,34 +46,51 @@ export function debugIdUploadPlugin({
...
-          (debugIdChunkFilePath) =>
-            debugIdChunkFilePath.endsWith(".js") ||
-            debugIdChunkFilePath.endsWith(".mjs") ||
-            debugIdChunkFilePath.endsWith(".cjs")
+      let globAssets;
+      if (assets) {
+        globAssets = assets;
+      } else {
+        logger.debug(
+          "Not `sourcemaps.assets` option provided, falling back to uploading detected build artifacts."
```
Lms24, line 67:
```suggestion
          "No `sourcemaps.assets` option provided, falling back to uploading detected build artifacts."
```

Lms24 (approved):

packages/bundler-plugin-core/src/debug-id-upload.ts
Lms24: Oh good point. At the time of writing I thought for some reason an empty array and an empty string the same semantic implications here but you're right, an empty string looks more like an error than an empty array.

Issue #260, opened 2023-05-23
Vue source map upload no longer working in 2.0


richardsimko:
### Environment

```
@sentry/vite-plugin@2.0.0
"@vitejs/plugin-vue": "4.2.3",
    "@vitejs/plugin-vue-jsx": "3.0.1",
"vite": "4.3.8",
```

### Steps to Reproduce
Config:
```ts
const prodConfig: UserConfig = {
  build: {
    manifest: true,
    sourcemap: true,
    emptyOutDir: true,
    chunkSizeWarningLimit: 600,
    target: "es2019",
    rollupOptions: {
      output: {
        inlineDynamicImports: false,
        dir: resolve(__dirname, "..", "public", "bundle"),
        manualChunks: {
          "error-handling": ["@sentry/vue"],
          vue: ["vue", "@vue/compiler-core", "@vue/compiler-dom"],
          lodash: ["lodash-es"],
          moment: ["moment"],
          popper: ["@popperjs/core"],
          consent: ["cookieconsent"],
          toastify: ["toastify-js"],
        },
      },
      input: "app/client/index.ts",
    },
  },
  resolve: {
    alias: {
      // '@': fileURLToPath(new URL('./src', import.meta.url)),
      mixins: fileURLToPath(new URL("./css/mixins.styl", import.meta.url)),
      vue: "vue/dist/vue.esm-bundler.js",
    },
  },
  plugins: [
    vue({
      isProduction: true,
      template: {
        transformAssetUrls: {
          includeAbsolute: false,
        },
      },
    }),
    vueJsx(),
    sentryVitePlugin({
      org: "org",
      project: "client",
      // release: process.env.COMMIT_HASH,
      release: { name: process.env.COMMIT_HASH },
      authToken: process.env.SENTRY_AUTH_TOKEN,
      sourcemaps: {
        // Specify the directory containing build artifacts
        assets: "./app/public/bundle/**/*",
        // deleteFilesAfterUpload: './app/public/bundle/**/*.map',
      },
    }),
  ],
};
```

### Expected Result

Source map upload works as it did in 0.7.2 (Need to change the `release` config attribute to match the breaking change)

### Actual Result

It seems Sentry is injecting some code in places where it shouldn't be. The file referenced is probably the first `.vue` file which gets loaded.

```
vite v4.3.8 building for production...
[...]

PR #270, opened 2023-05-23
fix: Use banner hooks for injection in vite and rollup

lforst:
How did I not think of this earlier...

Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/260

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -1,6 +1,5 @@
 import SentryCli from "@sentry/cli";
 import fs from "fs";
-import MagicString from "magic-string";
 import { createUnplugin, UnpluginOptions } from "unplugin";
 import { normalizeUserOptions, validateOptions } from "./options-mapping";
 import { debugIdUploadPlugin } from "./plugins/debug-id-upload";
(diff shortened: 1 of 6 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -15,9 +14,10 @@ import {
   getDependencies,
   getPackageJson,
   parseMajorVersion,
-  stringToUUID,
 } from "./utils";
 
+import { v4 as uuidV4 } from "uuid";
```
Lms24, line 19: l: What's our minimum supported Node version again? 😅 
If it is >14 we should be good to use [`randomUUID()`](https://stackoverflow.com/a/67624847), right? 
(but overall logaf-super-l)

packages/bundler-plugin-core/src/index.ts
```diff
@@ -253,90 +253,15 @@ export function sentryCliBinaryExists(): boolean {
...
-      // Appending instead of prepending has less probability of mucking with user's source maps.
-      // Luckily import statements get hoisted to the top anyways.
-      ms.append(`\n\n;import "${virtualReleaseInjectionFileId}";`);
-
-      return {
-        code: ms.toString(),
-        map: ms.generateMap(),
-      };
-    },
+    banner: injectionCode,
```
Lms24, line 257: Hmm wasn't there some requirement for us to be below `use strict;` or something like that? But other than that, good catch 😅

Issue #255, opened 2023-05-23
@sentry/webpack-plugin@2.0.0 installs its own copy of Webpack 4


wojtekmaj:
Installing the latest version of @sentry/webpack-plugin, 2.0.0, results in "webpack-4" dependency installed, along with ~250 other dependencies. This makes this plugin's install size [a whopping 51 MB](https://packagephobia.com/result?p=@sentry/webpack-plugin@2.0.0).

I understand there must be some reasoning behind this, and you probably spent much more time on this than I did, but this seems like a kind of workaround you should never be forced to do. I'm opening this thread in hope that we, as a community, could help out and figure something out to avoid this.

lforst:
Hi, thanks for raising this :) I am gonna give some context so that we can come to a solution faster. I would also like to improve this.

So we obviously want to support both, webpack 4 and webpack 5. For a particular set of features (release value injection and debug Id injection) we currently use webpack's `BannerPlugin`. It was the easiest and most robust way of being able to inject values (at the top of the bundle!) while keeping source maps intact. Building our own plugin for this that was compatible with Webpack 4 and 5 AND kept source maps intact was a nightmare - I actually couldn't get it to work.

The thing is, webpack 4 and webpack 5 ship their `BannerPlugin`s differently. Webpack 5 makes it accessible through `compiler.webpack.BannerPlugin` inside plugins, essentially making us get away without any webpack dependency at all, whereas with Webpack 4 you have to directly import it.

I don't quite remember what I tried out already but here's some stuff I wanna try out making webpack a peer dependency and importing the plugin directly for v5 and v4.

wojtekmaj:
Hmmm... Assuming `@sentry/webpack-plugin` can only be used with Webpack:

* `webpack: 4.x | 5.x` should be a peerDependency

And:

* For Webpack 5, we continue to use `compiler.webpack.BannerPlugin`, without the need of importing anything
* For Webpack 4, we could import`"webpack"` (which must be v4) provided to us by parent project

The problem is that:

```js
import { BannerPlugin as Webpack4BannerPlugin } from "webpack";
```

may crash when BannerPlugin is not exported, which I understand is the case with Webpack 5, hence the `… from "webpack-4"` trick.

But how about:

```js
import * as webpackMaybe4Exports from "webpack";
```

and then

```js
const BannerPlugin = compiler?.webpack?.BannerPlugin || webpackMaybe4Exports.BannerPlugin;

// …

[...]

PR #273, opened 2023-05-24
ref(webpack): Use webpack peer dependency

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/255

### DIFF
--- a/packages/webpack-plugin/package.json
+++ b/packages/webpack-plugin/package.json
@@ -50,8 +50,7 @@
   "dependencies": {
     "@sentry/bundler-plugin-core": "2.0.0",
     "unplugin": "1.0.1",
-    "uuid": "^9.0.0",
-    "webpack-4": "npm:webpack@^4"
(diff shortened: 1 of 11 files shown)


Lms24 (approved):

packages/webpack-plugin/package.json
```diff
@@ -66,12 +65,17 @@
...
     "jest": "^28.1.1",
     "rimraf": "^3.0.2",
     "rollup": "2.75.7",
     "ts-node": "^10.9.1",
-    "typescript": "^4.7.4"
+    "typescript": "^4.7.4",
+    "webpack": "npm:webpack@^4"
+  },
+  "peerDependencies": {
+    "webpack": ">=4"
```
wojtekmaj, line 78: That's a bit too optimistic. How about:

```suggestion
    "webpack": "^4.0.0 || ^5.0.0"
```
?

PR #290, opened 2023-05-26
fix(webpack): Set minimum webpack 4 peer dep to `4.40.0`

lforst:
Found problem in https://stackoverflow.com/a/76319670/21967320


AbhiPrasad (approved):
Shouldn't this be a fix? Since this is technically breaking for those who have `--strict-peer-deps` set.

lforst:
> Shouldn't this be a fix? Since this is technically breaking for those who have --strict-peer-deps set.

Ah true, good point. Thanks!

Issue #288, opened 2023-05-25
Track User Agent


scefali:
We should modify the user agent to say which bundler it came from so we can track it in analytics better. Currently, only the webpack bundler supports it.

lforst:
Hey, can you elaborate a bit? What do you mean with user agent? Do you refer to SDK event's user agent or the bundler telemetry?

scefali:
@lforst I'm talking about the bundler. Basically extend this: https://github.com/getsentry/action-release/blob/7d9c3a86a287efcbedca1328c56e15f2f15d5f80/src/cli.ts#L13

PR #296, opened 2023-05-31
feat(core): Set custom user agent when interacting with Sentry

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/288

Like we did in the old webpack plugin: https://github.com/getsentry/sentry-webpack-plugin/blob/3628137a10c93e0bf8f8362abc19c0b4c46218ca/src/index.js#L12

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -92,6 +92,11 @@ export function sentryUnpluginFactory({
       debug: options.debug,
     });
 
+    // Set the User-Agent that Sentry CLI will use when interacting with Sentry
+    process.env[
+      "SENTRY_PIPELINE"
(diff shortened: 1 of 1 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -92,6 +92,11 @@ export function sentryUnpluginFactory({
       debug: options.debug,
     });
 
+    // Set the User-Agent that Sentry CLI will use when interacting with Sentry
+    process.env[
```
mydea, line 96: super l, just wondering, is there a reason to not just do:

```
process.env.SENTRY_PIPLELINE = ...
```

mydea (approved):

packages/bundler-plugin-core/src/index.ts
lforst: its just lint rules

Issue #298, opened 2023-06-01
`ERROR: Unexpected ";"` when using sentryVitePlugin


MathiasWP:
### Environment

Using: `"@sentry/vite-plugin": "^2.2.0"`
Framework: `"@sveltejs/kit": "1.20.0"`

This is the config i'm using for the plugin:

```ts
sentryVitePlugin({
	org: 'my-org',
	project: 'my-project',
	authToken: env.SENTRY_AUTH_TOKEN,
	sourcemaps: {
		ignore: ['node_modules', 'vite.config.ts'],
		// Specify the directory containing build artifacts
		assets: './.svelte-kit/**'
	}
})
```

### Steps to Reproduce

1. Built my project

### Expected Result

Build should complete with no syntax error. This has happened for over a week now. We use dependabot, so a new build runs every time the plugin gets an update.

### Actual Result

It seems like Sentry is injecting some code that causes a syntax error in the transformed code.

```bash
rendering chunks (46)...[vite:esbuild-transpile] Transform failed with 1 error:
chunks/navigation.js:49:38: ERROR: Unexpected ";"

Unexpected ";"
47 |   * @returns {void}
48 |   */
49 |  const afterNavigate = /* @__PURE__ */ ;!function(){try{var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof self?self:{},n=(new Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="21ba7d35-8a1d-4997-1126-b7ce028af103",e._sentryDebugIdIdentifier="sentry-dbid-21ba7d35-8a1d-4997-1126-b7ce028af103")}catch(e){}}();client_method('after_navigate');
   |                                        ^
50 |  
51 |  export { afterNavigate as a, beforeNavigate as b };

✓ built in 11.84s
error during build:
Error: Transform failed with 1 error:
chunks/navigation.js:49:38: ERROR: Unexpected ";"
    at failureErrorWithLog (/tech/frontend/node_modules/esbuild/lib/main.js:1636:15)
    at /tech/frontend/node_modules/esbuild/lib/main.js:837:29
    at responseCallbacks.<computed> (/tech/frontend/node_modules/esbuild/lib/main.js:697:9)
    at handleIncomingPacket (/tech/frontend/node_modules/esbuild/lib/main.js:752:9)
[...]

PR #302, opened 2023-06-06
fix: Use intro hook for debug ID injection in rollup and vite

lforst:
Implicitly fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/298

I realized we were using an overkill approach of injecting something into each bundle. Rollup and vite have a `intro` hook that allows to inject code into bundles, similar to `banner` and `footer` except for inside any format-specific wrappers (and also below any JS directives). I was previously only looking at banner and footer which had the problem of being injected before JS directives - so I opted for the solution which we're getting rid of in this PR instead.

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -15,7 +15,6 @@ import {
   getDependencies,
   getPackageJson,
   parseMajorVersion,
-  stringToUUID,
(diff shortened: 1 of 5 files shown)


lforst:
Using the intro hooks will not allow us to have deterministic debug ids, potentially busting a lot of caches.

Issue #310, opened 2023-06-09
esbuild plugin causes errors in combination with other resolver plugins


lforst:
A customer reported

```text
ERROR node_modules/foo/bar/baz.css:40 : Cannot use "sentry-debug-id-proxy:foo/bar/baz-icons.woff?yr7byn" as a URL
```

I assume we are causing a problem because we are setting a namespace that cascades all the way down to user-code.

PR #311, opened 2023-06-09
fix(esbuild): Don't use namespace for esbuild proxy resolving

lforst:
Potentially fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/310

### DIFF
--- a/packages/esbuild-plugin/src/index.ts
+++ b/packages/esbuild-plugin/src/index.ts
@@ -38,7 +38,6 @@ function esbuildReleaseInjectionPlugin(injectionCode: string): UnpluginOptions {
 
 function esbuildDebugIdInjectionPlugin(): UnpluginOptions {
   const pluginName = "sentry-esbuild-debug-id-injection-plugin";
-  const proxyNamespace = "sentry-debug-id-proxy";
   const stubNamespace = "sentry-debug-id-stub";
 
   return {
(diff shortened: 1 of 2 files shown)


packages/esbuild-plugin/src/index.ts
```diff
@@ -38,34 +38,42 @@ function esbuildReleaseInjectionPlugin(injectionCode: string): UnpluginOptions {
...
-  const proxyNamespace = "sentry-debug-id-proxy";
   const stubNamespace = "sentry-debug-id-stub";
 
   return {
     name: pluginName,
 
     esbuild: {
       setup({ onLoad, onResolve }) {
-        onResolve({ filter: /.*/ }, (args) => {
+        onResolve({ filter: /.*/, namespace: "file" }, (args) => {
```
mydea, line 48: do we need this? We are just hard-coding `file` below 😅

packages/esbuild-plugin/src/index.ts
```diff
@@ -38,34 +38,42 @@ function esbuildReleaseInjectionPlugin(injectionCode: string): UnpluginOptions {
...
             pluginData: {
+              isProxyResolver: true,
               originalPath: args.path,
               originalResolveDir: args.resolveDir,
             },
           };
         });
 
-        onLoad({ filter: /.*/, namespace: proxyNamespace }, (args) => {
+        onLoad({ filter: /.*/, namespace: "file" }, (args) => {
```
mydea, line 66: also here, do we need `namespace` here?

packages/esbuild-plugin/src/index.ts
lforst: You are right, the namespace filter here is very likely unnecessary since we only want to resolve entrypoints which _should_ only have the "file" namespace anyways (since "file" is the default).

The "passthrough" namespace below is then also unnecessary --> removed in 5c24b303061dcfb5ab5e73db1eba518b81f948fd

packages/esbuild-plugin/src/index.ts
lforst: Yeah this namespace is also unnecessary. I think I defined it because I wanted to be conservative over what files we resolve/load, but thinking about it again, it seems unnecesary.

Thanks for questioning this!

lforst: 29f39f630cdb3de6865a6c627a1ec629e2b53b5f

mydea (approved):

PR #312, opened 2023-06-12
docs: Add explanation to migration docs why `RewriteFrames` shouldn't be used

lforst:
We see people having issues with debug IDs because they're still using the `RewriteFrames` integration from previous setups.

This PR adds a section to the migration docs clarifying that it should be removed when using debug IDs.


mydea (approved):
😬 it is not ideal that this is needed... is there a way we could automatically disable it if we detect debug ids are used? 🤔 but for now makes sense to add this at least.

lforst:
PR became redundant because we fixed this in the SDK https://github.com/getsentry/sentry-javascript/pull/8347

PR #318, opened 2023-06-14
feat: Sort globbed files to ensure deterministic bundle IDs

smbroadley:
The results from `glob()` are not guarenteed to be ordered, and prevent deterministic output.

This manifests in the assignment of artefeact names, as well as the bundle-id.

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -89,7 +89,9 @@ export function createDebugIdUploadFunction({
           debugIdChunkFilePath.endsWith(".cjs")
       );
 
-      debugIdChunkFilePaths.sort(); // ensure order; glob() does not
+      // The order of the files output by glob() is not deterministic
+      // Ensure order within the files so that {debug-id}-{chunkIndex} coupling is consistent
+      debugIdChunkFilePaths.sort();
 
       if (Array.isArray(assets) && assets.length === 0) {
         logger.debug(
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -89,6 +89,8 @@ export function createDebugIdUploadFunction({
           debugIdChunkFilePath.endsWith(".cjs")
       );
 
+      debugIdChunkFilePaths.sort(); // ensure order; glob() does not
+
       if (Array.isArray(assets) && assets.length === 0) {
         logger.debug(
           "Empty `sourcemaps.assets` option provided. Will not upload sourcemaps with debug ID."


lforst:
Hi, thanks for the contribution. I am struggling a bit to understand how sorting the globbing will in any way affect the determinism of the plugin since globbing doesn't affect the actual output of the plugins (apart from maybe what ends up inside an os tmp folder).

Artifact names are currently determined by the contained debug id, which are currently just random uuids. We haven't figured out a way of making this deterministic for all bundlers yet. We are open to suggestions.

Big hope lies in the bundlers themselves to emit debug IDs. [Bun](https://bun.sh/) already does this based on our input:

https://github.com/oven-sh/bun/blob/012a8ba3736a671e914171574eea9357f4f659f3/src/sourcemap/sourcemap.zig#L1369-L1384

We are actively pushing in TC39 for this to become a standard.

smbroadley:
Assume I have overriden the generation of random UUIDs (which I have ;)).

The order of those files affects the generated bundle-id, as without sorting, the `{debug-id}-{chunkIndex}` coupling is random. Sorting ensures we get predictable pairings, and stable Bundle-Ids.

lforst:
> The order of those files affects the generated bundle-id, as without sorting, the {debug-id}-{chunkIndex} coupling is random. Sorting ensures we get predictable pairings, and stable Bundle-Ids.

That makes sense. Thanks!

> Assume I have overriden the generation of random UUIDs (which I have ;)).

Just because I am interested: What does your approach look like?

lforst (approved):

smbroadley:
Please do not think bad of me. I'm just a guy trying to get deterministic builds whilst you sort it out proper, like. 😉 

```js
function stevesMegaHack(plugin) {
	return {
		apply(compiler) {
			console.log("Hijacking Sentry.... ***evil laugh***")
			plugin.apply({ // fake webpack compiler:
				webpack: compiler.webpack,
				hooks: compiler.hooks,
				options: {
					plugins: {
						push: (new_plugin)=>{
							if( new_plugin.constructor?.name === "BannerPlugin" 
							 && typeof(new_plugin.options.banner) === "function" ) { // we want this one
								new_plugin.banner = ({chunk})=>{
									const it = (function*(s){while(1)yield* s})(chunk.hash)
									const guid = 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
											     .replace(/x/g, x=>it.next().value)

									return bundlerPluginCore.getDebugIdSnippet(guid);
								}
							}

							compiler.options.plugins.push(new_plugin)
						}
					}
				}
			})
		},
	}
}
```
...and then in my webpack.config.js:
```js
	plugins: [
		stevesMegaHack(
			sentryWebpackPlugin({ ... })
		)
	]
```

lforst:
> Please do not think bad of me.

@smbroadley Nah man that is hella smart we can add that to the plugin!

smbroadley:
That's great news - Thank you, Luca!

PR #321, opened 2023-06-15
feat(webpack): Generate deterministic debug IDs

lforst:
Use a hash argument provided by the webpack `BannerPlugin` in  order to deterministically generate a debug ID.

cc @smbroadley


packages/webpack-plugin/src/index.ts
```diff
@@ -42,7 +53,10 @@ function webpackDebugIdInjectionPlugin(): UnpluginOptions {
         new BannerPlugin({
           raw: true,
           include: /\.(js|ts|jsx|tsx|mjs|cjs)$/,
-          banner: () => getDebugIdSnippet(uuidv4()),
+          banner: (arg?: BannerPluginCallbackArg) => {
+            const debugId = arg?.chunk?.hash ? stringToUUID(arg?.chunk?.hash) : uuidv4();
```
AbhiPrasad, line 57: when is the hash not available? 

If so is this even truely deterministic? Is there another string we can fallback on if the hash does not exist?

can we get a test to confirm?

lforst: From the type def:

Screenshot 2023-06-15 at 16 53 54

I honestly more or less did this out of caution. Generally, it should be available though.

> If so is this even truely deterministic?

Should be, yes.

>  Is there another string we can fallback on if the hash does not exist?

I didn't investigate yet. For now I thought this was a quick win that works 99% of the time.

lforst: Test I will add!

packages/integration-tests/fixtures/deterministic-debug-ids/deterministic-debug-ids.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    path.join(__dirname, "input", "dynamic-variable.js"),
+    `global.dynamicVariable = 1;`
+  );
+});
+
+afterAll(async () => {
+  await fs.unlink(path.join(__dirname, "input", "dynamic-variable.js"));
+});
+
+test("Same debug IDs for multiple identical builds", () => {
```
AbhiPrasad, line 31: l: This should either be a table test, or separate tests for rollup/vite/webpack.

Makes it easier to add new tests in future, and to instantly identify which bundler is broken

AbhiPrasad (approved):

Issue #326, opened 2023-06-20
Getting Syntax error probably due to the optional chaining.


ukhanzada-nisum-com:
### Environment
"@sentry/react": "^7.52.1",
"@sentry/webpack-plugin": "^2.3.0",
"webpack": "^4.44.2",
"webpack-cli": "^3.3.12",
"node": "12.0.0"

What version are you running? Etc.
<img width="877" alt="Screen Shot 2023-06-15 at 5 28 09 PM" src="https://github.com/getsentry/sentry-javascript-bundler-plugins/assets/38420106/0e2bfe67-0a8f-4cf8-ac0f-424914ce36a7">

### Steps to Reproduce

1. When I'm running the project getting this error

### Expected Result

What you thought would happen.

### Actual Result

What actually happened. Maybe a screenshot/recording? Maybe some logs?

lforst:
Hi, the minimum supported node version of the plugin is Node.js 14. We need to update the engines fields!

PR #327, opened 2023-06-20
docs: Update minimum supported Node.js version to 14

lforst:
Ref https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/326

The unplugin dependency uses optional chaining which is only available in Node 14 or higher. Additionally, we only run our tests for versions 14 and higher.


mydea (approved):
not sure if this _technically_ is a breaking change, but I guess even before it would not really have worked with node < 14? So I guess this is fine!

lforst:
> not sure if this technically is a breaking change, but I guess even before it would not really have worked with node < 14? So I guess this is fine!

Yeah this change was considered as part of the v2 major bump. I simply forgot to document it.

Issue #332, opened 2023-06-27
Add a way to configure the plugins via an external file


lforst:
In `v2` of the bundler plugins we got rid of the functionality to configure the plugins with a [`.sentryclirc` file](https://docs.sentry.io/product/cli/configuration/#to-authenticate-manually). We wanted to abstract away as much as possible that Sentry CLI is used under the hood. It allowed us to reduce complexity and be way more specific in the error messages we print.

When working on a [source map CLI wizard](https://github.com/getsentry/sentry-wizard/issues/290), we noticed that we need a file that we can `gitignore` and put an auth token into for a quick setup.

We thought about bringing back `.sentryclirc` but that would leave us with the drawbacks we steered away from in `v2`. So now we decided to go with a new file that aims to be a bit more clear in what it does when paired with the bundler plugins:

`.env.sentry-build-plugin`

We settled for the `.env` file format because it is the config format that JS devs are most used to besides `.json`. We explored other names for the file:

- `.env.sentry` (too generic)
- `.env.sentry-vite-plugin` (complexity across bundlers, also complicated to document)
- `.env.sentry-plugin` (might be confused with SDK configuration)
- `.env.sentry-ci` (plugins can run locally too, not just in CI)
- `.env.sentry-auth` (too specific - we could have other things in that config too)
- `.env.sentry-sourcemaps` (too specific - it's not just about source maps)

---

The plan is to just dump all environment variables inside `.env.sentry-build-plugin` onto `process.env` when a plugin is initialized.

We also want to print a log message when a `.env.sentry-build-plugin` file is loaded.

For now, we only want to look for the file in `process.cwd()`. We thought about doing the discovery with an upwards folder search but that comes with complications, like how do you disable searching upwards when you don't want to use a higher nested config etc.

We also thought about adding something like an `envFile` option. For now, we don't see it as necessary, as alternative people can just provide all the values that would be inside `.env.sentry-build-plugin` manually as options. The main use case for this functionality is the source map wizard.

nemoDreamer:
> We wanted to abstract away as much as possible that Sentry CLI is used under the hood.

But why, tho?

Since the Sentry UI tells folks to run `sentry-cli sourcemaps ...` to debug their issues, we anyway have to have a `.sentryclirc` locally.

Now I have to provide our config using 2 different mechanisms, on the same machine... 😭

lforst:
> But why, tho?

[...]

PR #333, opened 2023-06-27
feat: Add configuration via `.env.sentry-build-plugin` file

lforst:
Resolves https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/332

Adds option to configure environment variables via a `.env.sentry-build-plugin` file inside the cwd.

This is necessary for efforts in https://github.com/getsentry/sentry-wizard/issues/290

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -69,7 +69,12 @@ export function sentryUnpluginFactory({
       path: path.join(process.cwd(), ".env.sentry-build-plugin"),
     });
 
-    if ("parsed" in dotenvResult) {
+    // Ignore "file not found" errors but throw all others
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Accessing `code` on error should be safe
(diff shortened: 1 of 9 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -58,6 +59,20 @@ export function sentryUnpluginFactory({
...
+      prefix: `[sentry-${unpluginMetaContext.framework}-plugin]`,
+      silent: userOptions.silent ?? false,
+      debug: userOptions.debug ?? false,
+    });
+
+    const dotenvResult = dotenv.config({
+      path: path.join(process.cwd(), ".env.sentry-build-plugin"),
+    });
+
+    if ("parsed" in dotenvResult) {
```
Lms24, line 72: l/q: What happens if this errored? Should we log something in this case? Just wondering if in an error case, some variables might have been pulled in while others weren't?...

Lms24 (approved):
Nice, LGTM!

packages/bundler-plugin-core/src/index.ts
lforst: Good point. Added that we throw on any other error except for ENOENT!

PR #334, opened 2023-06-28
feat: Add module metadata injection

### LINKED ISSUE #323: [webpack & MFEs] Able to uniquely identified JS code associated with different teams
### Background
Micro frontends are inherently ambiguous outside of certain frameworks for MFE delivery (module federation for example). Ultimately users want to be able to keep their code and errors separated, to be able to understand where regressions might have occurred specifically and functionally to be able to route errors within their teams.

### Proposal
With our webpack plugin it is possible during build time to add identifiers to the generated/built code. For example the `debugId`s added in [Artifact Bundles](https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/artifact-bundles/).

Our webpack plugin similarly could inject unique id for code which should be respected as part of their own verticals. The SDK at runtime would then have access to these `containerId`s and could then be handled in beforeSend or similar, to route events generated to a specific DSN, add tags to be able to filter in the Sentry UI and more

- [ ]  can use injected IDs to populate events with information maintained by bundler-plugin to store on events as context to send to sentry
- [ ]  SDK will not make any automatic decision at this stage of routing (projects, teams, multiple projects or orgs etc)
- [ ]  ensure the data is populated in Sentry and consumable via API

**Consideration:**
This work will be restricted to only webpack plugin for now, but it is important to note that something like this could be applied outside of webpack, and inversely more specifically focus on Module Federation concepts. Then further if it might be useful one day to have this upstream in Module Federation, webpack generically and other bundlers.

For now I think it would be the best interest of ourselves and users to try and conform namings we choose to match those used withing webpack Module Federation to make it easy to understand for casual and expert users alike.

timfish:
Closes #323

This PR:
- Adds `moduleMetadata` to `_experiments` option
- Adds an optional plugin for injecting module metadata
- Only adds plugin when it exists and `moduleMetadata` has been set
- Adds an implementation for the webpack plugin 

This PR does not yet pass through `entryPoint` to the metadata callback as [suggested here](https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/323#issuecomment-1607308176).

- Needs tests!


packages/bundler-plugin-core/src/types.ts
```diff
@@ -241,9 +241,26 @@ export interface Options {
...
+     */
+    moduleMetadata?: object | ModuleMetadataCallback;
   };
 }
 
+export interface ModuleMetadataCallbackArgs {
+  org?: string;
+  project?: string;
+  release?: string;
+  dist?: string;
```
lforst, line 259: I don't think we need dist as callback arg since it is never generated by the plugin itself.

timfish:
The newly added e2e tests pass on my machine...

lforst:
> The newly added e2e tests pass on my machine but don't run in CI from my fork...

Ah we disabled E2E tests since we introduced debug IDs since the ones we had didn't make sense anymore.

Lemme look into this...

timfish:
> we disabled E2E tests since we introduced debug IDs since the ones we had didn't make sense anymore

My assumption was that e2e tests are disabled for PRs from external forks since an auth token is used for the `basic_upload` test.

lforst:
> > we disabled E2E tests since we introduced debug IDs since the ones we had didn't make sense anymore
> 
> My assumption was that e2e tests are disabled for PRs from external forks since an auth token is used for the `basic_upload` test.

That too!

packages/bundler-plugin-core/src/utils.ts
```diff
@@ -264,6 +264,24 @@ export function generateGlobalInjectorCode({
...
+      typeof window !== 'undefined' ?
+        window :
+        typeof global !== 'undefined' ?
+          global :
+          typeof self !== 'undefined' ?
+            self :
+            {};
+
+    _global2.__MODULE_METADATA__ = _global2.__MODULE_METADATA__ || {};
+    _global2.__MODULE_METADATA__[new Error().stack] = ${JSON.stringify(metadata)};`;
```
lforst, line 282: https://github.com/getsentry/sentry-javascript/pull/8438#discussion_r1250736786

In case we change this property name we should align these two PRs.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -241,9 +241,25 @@ export interface Options {
...
   };
 }
 
+export interface ModuleMetadataCallbackArgs {
+  org?: string;
+  project?: string;
+  release?: string;
+}
+
+export type ModuleMetadataCallback = (args: ModuleMetadataCallbackArgs) => object;
```
lforst, line 261: Should we set the return type of this function to `any`? Users are allowed to return anything they want here in theory.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -241,9 +241,25 @@ export interface Options {
...
      */
     injectBuildInformation?: boolean;
+
+    /**
+     * Metadata associated with this module.
+     * The metadata is serialized and can be looked up at runtime by URL.
+     *
+     * Note: This option is currently only supported by `@sentry/webpack-plugin`.
+     */
+    moduleMetadata?: object | ModuleMetadataCallback;
```
lforst, line 251: Hm. I would like to make this `any | ModuleMetadataCallback` but it's just gonna get turned into `any`.

I think we should just do it and people who follow the definition will see that this also accepts a callback. We should add to the jsdoc that this can be a static value or a callback.

packages/e2e-tests/scenarios/module-metadata/module-metadata.test.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from "path";
```
lforst, line 1: Can we translate this E2E test into a integration test instead.

IIRC the E2E tests were built for querying the Sentry API for uploaded artifacts.

packages/bundler-plugin-core/src/types.ts
timfish: Yeah I think that's probably a good idea

packages/bundler-plugin-core/src/types.ts
timfish:
```ts
type MetadataValue = object | boolean | number | string | symbol`;

moduleMetadata?: MetadataValue | ModuleMetadataCallback;
```
😆

packages/e2e-tests/scenarios/module-metadata/module-metadata.test.ts
timfish: I completely missed the fact there are separate integration tests 🙈

lforst (approved):
Nice, thank you! :)

Issue #338, opened 2023-07-06
Conflicting information about sourcemap upload


oliversalzburg:
<https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/main/packages/bundler-plugin-core/src/types.ts#L88-L91> says:

```js
  /**
   * Options for source maps uploading.
   * Leave this option undefined if you do not want to upload source maps to Sentry.
   */
```

This seems to indicate pretty clearly that sourcemap will not be uploaded if `sourcemaps` is left `undefined`.

The documentation at <https://docs.sentry.io/platforms/node/sourcemaps/uploading/esbuild> seems to ignore this, and sets no `sourcemaps`, which would prevent sourcemap uploading, which seems counter-intuitive, given that esbuild was instructed with `sourcemap: true` here.

Which is it?

I'm trying to troubleshoot why I'm not seeing source code for stack traces on Sentry and this conflicting information makes it hard to determine if anything is working as intended.

lforst:
Ah, you're right, this comment is a leftover from a previous version. We should remove it.

PR #339, opened 2023-07-06
docs: Remove misleading documentation

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/338

We previously documented that leaving away the `sourcemaps` option will disable debug ID injection. That is pre v2 behaviour and not the case anymore so we should remove it.

### DIFF
--- a/packages/dev-utils/src/generate-documentation-table.ts
+++ b/packages/dev-utils/src/generate-documentation-table.ts
@@ -72,8 +72,7 @@ errorHandler: (err) => {
   },
   {
     name: "sourcemaps",
-    fullDescription:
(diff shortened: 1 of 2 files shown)


packages/dev-utils/src/generate-documentation-table.ts
```diff
@@ -72,8 +72,7 @@ errorHandler: (err) => {
   },
   {
     name: "sourcemaps",
-    fullDescription:
-      "Options for source maps uploading. Leave this option undefined if you do not want to upload source maps to Sentry.",
+    fullDescription: "Options for source maps uploading.",
```
Lms24, line 75: super-l:
```suggestion
    fullDescription: "Options for uploading source maps.",
```

Lms24 (approved):

PR #355, opened 2023-07-28
deps: Bump sentry-cli to 2.20.1 to fix debugid injection issue

kamilogorek:
Fixes https://github.com/getsentry/sentry-javascript/issues/8665


lforst:
We're not using the cli to inject debug id snippets but it cant hurt in case the pinned version is picked up somehow

lforst (approved):

PR #358, opened 2023-07-31
fix: Support `.jsbundle`s in debug ID uploads

raon0211:
React Native bundles usually use `*.jsbundle` as their extensions, but Sentry is not supporting this extension when uploading debug IDs. Added a line to support `*.jsbundle` extension.


lforst:
Hi, this generally looks good to me, however, I don't have much experience with react native. Did you try this out?

@krystofwoldrich I would be interested in your thoughts on this!

sennhann29 (approved):

krystofwoldrich:
@raon0211 @sennhann29 How does your setup look like are you using https://re-pack.netlify.app/ with the Sentry Webpack plugin?

I would add the `jsbundle` extension support only when we know the use cases when this is used and that it works both for JSC and Hermes.

Issue #365, opened 2023-08-07
Allow omitting `org` when using org auth tokens


mydea:
When a user provides an Org Auth Token (detectable by `sntrys_` prefix), and we are on latest sentry-cli (should bump it - supported since https://github.com/getsentry/sentry-cli/pull/1673), users can omit the `org`. Project is still required. No need to do anything else as sentry-cli will do the extraction of data, just allowing to not put an org in is all that should be needed here.

PR #368, opened 2023-08-08
feat: Allow ommiting `org` when using organization auth token

lforst:
Resolves https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/365


mydea (approved):
nice!

Issue #370, opened 2023-08-08
Webpack plugin `filesToDeleteAfterUpload`: ENOENT: no such file or directory


michaelbridge:
### Environment

@sentry/webpack-plugin
Version 2.5.0

### Steps to Reproduce

```
sentryWebpackPlugin({
	authToken: process.env.SENTRY_AUTH_TOKEN,
	errorHandler: err => {
		console.warn(err)
	},
	org: ...,
	project: ...,
	release: {
		name: process.env.SOURCE_VERSION,
	},
	sourcemaps: {
		assets: ['./dist/scripts/*.js', './dist/scripts/*.js.map'],
		filesToDeleteAfterUpload: './dist/scripts/*.js.map',
	},
	url: 'https://sentry.io/',
})
```

### Expected Result

Source maps uploaded; build completes.

### Actual Result

Source maps uploaded; build fails while deleting source maps.

```
sentry-webpack-plugin] Warning: Didn't find any matching sources for debug ID upload. Please check the `sourcemaps.assets` option.
       > Found 34 files
       > Analyzing 34 sources
       > Adding source map references
       > Bundled 34 files for upload
       > Bundle ID: ca236be7-c41c-59ef-a4ca-155f28d49516
       > Uploaded files to Sentry
       > File upload complete (processing pending on server)
       > Organization: 
       > Project: 
       > Release: 39f934505fd83398d619d47b2afed3531e1cdba5
       > Dist: None
       > Upload type: artifact bundle
       
       Source Map Upload Report
         Minified Scripts
           ~/079a5758-1ea3-4d97-bb45-d0fd688cb541-15.js (sourcemap at vendors~developer.f34fddbe540c96868a0e.js.map, debug id 079a5758-1ea3-4d97-bb45-d0fd688cb541)
           ~/1b452442-291f-4a9c-aa2e-a528ae6f0093-5.js (sourcemap at 14.a1c06fdd498543703ce5.js.map, debug id 1b452442-291f-4a9c-aa2e-a528ae6f0093)
           ~/1c197787-c2ba-4114-8444-0ca73c01b5bf-9.js (sourcemap at 3.9d8da54e1a41e67975a4.js.map, debug id 1c197787-c2ba-4114-8444-0ca73c01b5bf)
[...]

PR #373, opened 2023-08-09
fix: Don't crash on failed delete after upload

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/370


kamilogorek (approved):

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -181,7 +181,13 @@ export function createDebugIdUploadFunction({
         });
         await Promise.all(
           filePathsToDelete.map((filePathToDelete) =>
-            fs.promises.rm(filePathToDelete, { force: true })
```
kamilogorek, line 184: Or just use `Promise.allSettled()` everywhere. Then `catch` is redundant.

Issue #377, opened 2023-08-10
determineSourceMapPathFromBundle should not treat URLs as paths


tomyam1:
Following the fix for https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/375

I think my comment https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/375#issuecomment-1673185672 about the issue this introduced might have beeen missed

The current code will fail e.g. for

```
//# sourceMappingURL=https://path/to/bundle.js.map
```

tomyam1:
Thanks @lforst !

PR #378, opened 2023-08-10
fix: Make sourceMappingURL heuristic more resilient

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/377

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -298,9 +298,17 @@ async function determineSourceMapPathFromBundle(
   const sourceMappingUrlMatch = bundleSource.match(/^\s*\/\/# sourceMappingURL=(.*)$/m);
   if (sourceMappingUrlMatch) {
     const sourceMappingUrl = path.normalize(sourceMappingUrlMatch[1] as string);
-    let absoluteSourceMapPath;
 
-    if (path.isAbsolute(sourceMappingUrl)) {
+    let isSupportedUrl;
+    try {
+      const url = new URL(sourceMappingUrl);
+      isSupportedUrl = url.protocol === "file:";
+    } catch {
+      isSupportedUrl = false;
+    }
+
+    let absoluteSourceMapPath;
+    if (isSupportedUrl || path.isAbsolute(sourceMappingUrl)) {
       absoluteSourceMapPath = sourceMappingUrl;
     } else {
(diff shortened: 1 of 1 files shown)


packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -298,10 +298,28 @@ async function determineSourceMapPathFromBundle(
...
+    if (isSupportedUrl || path.isAbsolute(sourceMappingUrl)) {
+      absoluteSourceMapPath = sourceMappingUrl;
     } else {
-      return path.join(path.dirname(bundlePath), sourceMappingUrl);
+      absoluteSourceMapPath = path.join(path.dirname(bundlePath), sourceMappingUrl);
+    }
+
+    try {
+      // Check if the file actually exists
+      await util.promisify(fs.access)(absoluteSourceMapPath);
```
Lms24, line 319: l: any reason we use this over `fs.promises.access`?

(don't feel obligated to change it, I'm primarily just curious)

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -298,10 +298,28 @@ async function determineSourceMapPathFromBundle(
...
+      isSupportedUrl = false;
+    }
+
+    let absoluteSourceMapPath;
+    if (isSupportedUrl || path.isAbsolute(sourceMappingUrl)) {
+      absoluteSourceMapPath = sourceMappingUrl;
     } else {
-      return path.join(path.dirname(bundlePath), sourceMappingUrl);
+      absoluteSourceMapPath = path.join(path.dirname(bundlePath), sourceMappingUrl);
+    }
```
Lms24, line 315: m: We reach this branch even if `isSupportedUrl === false`. Is this by intentional?

packages/bundler-plugin-core/src/debug-id-upload.ts
lforst: No not really. I think originally I used it because it was available in earlier node versions I believe.

packages/bundler-plugin-core/src/debug-id-upload.ts
lforst: You're right! Fixed in 9c5b99fdd3a9b8273682d8d113528fb3eedcd032

Lms24 (approved):

PR #380, opened 2023-08-14
feat: Add module metadata injection for vite and rollup

timfish:
Metadata injection has already been added for webpack (#334).

This PR adds this feature for rollup/vite too.


lforst (approved):
Nice. Thanks!

Issue #351, opened 2023-07-20
Esbuild plugin does up inject/upload all source and map files on build


Fwang36:
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Which SDK are you using?

@sentry/browser

### SDK Version

@sentry/esbuild-plugin 2.4.0

### Framework Version

react 18.2

### Steps to Reproduce

1. Enable splitting in esbuild config with `splitting: true`
2. Create a dynamic import
3. Add the sentryEsbuildPlugin()
4. build

My esbuild config - 
![Screenshot 2023-07-19 at 10 06 24 AM](https://github.com/getsentry/sentry-javascript/assets/96670625/958dfe52-40ed-45fc-a2a2-1b4df3787e9b)

My output files - 
<img width="234" alt="Screenshot 2023-07-19 at 10 06 39 AM" src="https://github.com/getsentry/sentry-javascript/assets/96670625/49e7af64-efd3-4c37-9a1f-65b9e317b15d">

Upload logs where only app.js and app.js.map being uploaded - 
<img width="864" alt="Screenshot 2023-07-19 at 10 07 09 AM" src="https://github.com/getsentry/sentry-javascript/assets/96670625/5fb778af-e997-4898-924c-93a5b4352fb9">

### Expected Result

All the files outputted by the bundler should be injected and uploaded

### Actual Result

Not all files outputted by the bundler are being injected and uploaded.

lforst:
Took a look at this. Unfortunately, due to the restrictive nature of the esbuild plugin API we will probably not get debug ID injection working in combination with `splitting: true`. In these cases we will probably need to use `options.uploadLegacySourcemaps`.

lforst:
Putting in backlog and keeping an eye on the esbuild API. Maybe in the future this will be possible.

muuvmuuv:
Hey, I don't know if this is related or not, but when using the new Angular with esbuild and inject sentry esbuild plugin I see source maps in `buildArtifactPaths` but after `glob(globAssets)` all source maps get lost. Is that because the artifacts are not yet written?

lforst:
@muuvmuuv did you provide an `assets` option?

[...]

PR #381, opened 2023-08-15
feat: Add module metadata injection for esbuild

timfish:
Metadata injection has already been added for webpack (https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/334) and there is an open PR for rollup (#380).

This PR adds this feature for esbuild too which is mostly a copy from `esbuildDebugIdInjectionPlugin`.

My only minor concern is that due to `if (args.kind !== "entry-point") { return }`, it means that code is only injected for entry points. What happens for other chunks that aren't entry, like those created from async import, etc? 

I've done a bit of digging and it appears that esbuild doesn't yet support async import so maybe my concerns can be ignored for now.

### DIFF
--- a/packages/esbuild-plugin/src/index.ts
+++ b/packages/esbuild-plugin/src/index.ts
@@ -133,7 +133,7 @@ function esbuildModuleMetadataInjectionPlugin(injectionCode: string): UnpluginOp
               // needs to be an abs path, otherwise esbuild will complain
               path: path.isAbsolute(args.path) ? args.path : path.join(args.resolveDir, args.path),
               pluginData: {
-                isProxyResolver: true,
+                isMetadataProxyResolver: true,
                 originalPath: args.path,
                 originalResolveDir: args.resolveDir,
               },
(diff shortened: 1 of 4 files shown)


packages/esbuild-plugin/src/index.ts
```diff
@@ -115,6 +115,88 @@ function esbuildDebugIdInjectionPlugin(): UnpluginOptions {
...
+        onResolve({ filter: /.*/ }, (args) => {
+          if (args.kind !== "entry-point") {
+            return;
+          } else {
+            return {
+              pluginName,
+              // needs to be an abs path, otherwise esbuild will complain
+              path: path.isAbsolute(args.path) ? args.path : path.join(args.resolveDir, args.path),
+              pluginData: {
+                isProxyResolver: true,
```
lforst, line 136: I am thinking that we should probably chose a slightly different name for the flag here so we avoid a collision with the debug ID injection plugin.

Same goes for the `sentryProxyModule` suffix.

lforst (approved):
Nice. Thank you Tim!

As for

> My only minor concern is that due to if (args.kind !== "entry-point") { return }, it means that code is only injected for entry points. What happens for other chunks that aren't entry, like those created from async import, etc?
> 
> I've done a bit of digging and it appears that esbuild doesn't yet support async import so maybe my concerns can be ignored for now.

This is true. We also have an issue with `splitting: true` which is somewhat related: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/351

We may need to think about adding `dynamic-import` to the `args.kind` check in this if-statement.

PR #390, opened 2023-08-27
fix: Save results of rewriteSourcesHook

adonskoy:
`map()` returns a new array, doesn't overwrite the current one.


lforst (approved):
Thanks!

Issue #396, opened 2023-09-04
Fails with token error with dryRun enabled


rchl:
### Environment

@sentry/webpack-plugin@2.7.1

### Steps to Reproduce

Used the `@sentry/webpack-plugin` plugin with the following options:

```
{
      authToken: 'fakeToken',
      org: 'MyCompany',
      project: 'TestProject',
      dryRun: true,
      debug: true,
    }
```

### Expected Result

No exception thrown due to missing `authToken`.
No API requests to publish releases should be made.

### Actual Result

Getting unhandled exception:

```
Error: Command failed: /usr/local/workspace/nuxt/sentry-module/node_modules/@sentry/cli/sentry-cli releases new d5d66ef65ec55052dc94fe6965a2604e51791375
error: API request failed
  caused by: sentry reported an error: Invalid token (http status: 401)
```

rchl:
Thanks! Makes sense then.

PR #397, opened 2023-09-05
docs: Add removal of `dryRun` to migration guide

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/396


umanghome:
@lforst @Lms24 Is there an alternative to `dryRun` that is supported now?

Issue #413, opened 2023-10-02
`esbuild-plugin` overrides injected variables


ilijapuaca:
### Environment

What version are you running? Etc.

`@sentry/esbuild-plugin@2.7.1`, `esbuild@0.19.4`

### Steps to Reproduce

This surfaced after upgrading to esbuild `0.19.0`, but perhaps more widely spread and related to [this issue](https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/367)?

When `@sentry/esbuild-plugin` is used, the built bundle would crash with `ReferenceError: process is not defined` when run. I spent some time comparing the output with and without Sentry plugin, here's the minimal repro setup:

`build-config/build.js`
```
import {sentryEsbuildPlugin} from '@sentry/esbuild-plugin';
import esbuild from 'esbuild';
import path from 'path';

const sentryPlugin = sentryEsbuildPlugin({
  org: 'foo',
  project: 'bar',
  authToken: 'token'
});

esbuild.build({
  bundle: true,
  entryPoints: [path.resolve(__dirname, '../src/index.ts')],
  outdir: path.resolve(__dirname, '../dist'),
  inject: [path.resolve(__dirname, './inject.ts')],
  plugins: [sentryPlugin],
  minify: false,
  sourcemap: true
});
```

`build-config/inject.js`
```
export const process = {cwd: () => '', env: {}};
export const global = globalThis;
```

`src/index.ts`
```
const FOO = process.env.FOO ? 'foo' : 'bar';
console.log(FOO);
```

### Expected Result

Before esbuild `0.19` the output injection piece would look something like this:

```
...

  // build-config/inject.ts?sentryProxyModule=true
  var process, global;
  var init_inject = __esm({
    "build-config/inject.ts?sentryProxyModule=true"() {
      process = { cwd: () => "", env: {} };
      global = globalThis;
    }
  });

  // src/index.ts
  var require_src = __commonJS({
    "src/index.ts"() {
      init_inject();
      var FOO = process.env.FOO ? "foo" : "bar";
      console.log(FOO);
    }
  });

...
```

### Actual Result

Instead, now it outputs something like this:

```
...

  // sentry-debug-id-stub:_sentry-debug-id-injection-stub?sentry-module-id=xxx
  var init_sentry_debug_id_injection_stub = __esm({
    "sentry-debug-id-stub:_sentry-debug-id-injection-stub?sentry-module-id=xxx"() {
      init_inject2();
[...]

PR #416, opened 2023-10-05
test(esbuild): Add test to check for compatibility with `inject` option

lforst:
Ref: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/413

Verify that we are not stomping the `inject` option somehow.

### DIFF
--- a/packages/integration-tests/fixtures/esbuild-inject-compat/input/inject.ts
+++ b/packages/integration-tests/fixtures/esbuild-inject-compat/input/inject.ts
@@ -1,0 +1,3 @@
+export const process = { cwd: () => "", env: {} };
+// eslint-disable-next-line no-undef
+export const global = globalThis;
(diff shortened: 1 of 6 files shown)


packages/integration-tests/fixtures/esbuild-inject-compat/input/inject.ts
```diff
@@ -0,0 +1,3 @@
+export const process = { cwd: () => "", env: {} };
```
mydea, line 1: Is the idea that this _overwrites_ `process.env.FOO`? Where would that be set?

lforst: I could make the test a bit clearer actually...

lforst: Adjusted here [d32130e](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/416/commits/d32130e5ad6bac1f6b64d7ae6b681eb40cadefb8)

mydea (approved):

lforst:
Whoops this fails. But this is fixed with https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/417 🤦

PR #422, opened 2023-10-10
fix(core): Widen detection of source maps with `.cjs.map` and `.mjs.map`

lforst:


Lms24 (approved):
Thanks for taking care of this so quickly!

Issue #425, opened 2023-10-18
Add options for treeshaking through build-time flags


mydea:
Based on this https://docs.sentry.io/platforms/javascript/configuration/tree-shaking/, we should add flags to the bundler plugins to automatically shake out code you don't care about, something like:

```js
sentryWebpackPlugin({
  // other options...
  stripDebugStatements: true,
  stripRrwebCanvas: true,
  stripRrwebIframe: true,
  stripRrwebShadowDom: true
});
```

(naming TBD).

```[tasklist]
### Tasks
- [ ] Allow to strip debug statements (`__SENTRY_DEBUG__: false`)
- [ ] Allow to strip rrweb canvas capturing (`__RRWEB_EXCLUDE_CANVAS__`: true)
- [ ] Allow to strip rrweb iframe capturing (`__RRWEB_EXCLUDE_IFRAME__`: true)
- [ ] Allow to strip rrweb shadow dom capturing (`__RRWEB_EXCLUDE_SHADOW_DOM__`: true)
```

lforst:
Bikeshed:
- ~stripDebugStatements~ disableDebugStatements
- ~stripRrwebCanvas~ disableReplayCanvasSupport
- ~stripRrwebIframe~ disableReplayIframeSupport
- ~stripRrwebShadowDom~ disableReplayShadowDomSupport

Reasons:
- We cannot name it "strip" because we cannot control what will be treeshaken. Users can still acces the APIs preventing them from being shaken or they can disable tree shaking in their bundlers.
- We shouldn't name it "rrweb" because it is not well known.

mydea:
After talking a bit about this offline with Luca, let's put this into an object to make it clear what it does, e.g.:

```js
sentryWebpackPlugin({
  // other options...
  bundleSizeOptimizations: {
    excludeDebugStatements: true,
    excludeReplayCanvas: true,
    excludeReplayShadowDom: true,
    excludeReplayIframe: true
  }
});
```

I kind of prefer `exclude` or something like this over `disable`, but not strongly, so 🤷

PR #428, opened 2023-10-19
feat: Allow to configure `bundleSizeOptimizations`

mydea:
This introduces a new option for the plugins:

```
sentryPlugin({
   bundleSizeOptimizations: {
      excludeDebugStatements: true,
      excludePerformanceMonitoring: true,
      excludeReplayCanvas: true,
      excludeReplayIframe: true,
      excludeReplayShadowDom: true,
    },
});
```

When set, this will replace the relevant build time flags with a boolean, allow bundlers to treeshake code out and reduce the shipped bundle size.

Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/425


packages/bundler-plugin-core/src/index.ts
```diff
@@ -371,6 +399,18 @@ export function createRollupReleaseInjectionHooks(injectionCode: string) {
...
+export function createRollupBundleSizeOptimizationHooks(values: SentrySDKBuildFlags) {
+  return {
+    renderChunk(code: string) {
+      return replaceBooleanFlagsInCode(code, values);
+    },
+
+    transform(code: string) {
+      return replaceBooleanFlagsInCode(code, values);
+    },
+  };
```
Lms24, line 411: Why do we need to call the replace function in two hooks? Shouldn't `transform` be enough?

Lms24 (commented):
I like the API for this! Probably easier to use for the majority of folks than defining a replace plugin themselves. 

Do you think there's a way that we could filter out files before reading them? I'm thinking about looking at the file path and checking for a `@sentry` occurance. But not sure if I'm missing something tough why this wouldn't work.

packages/bundler-plugin-core/src/utils.ts
```diff
@@ -307,3 +308,27 @@ export function stripQueryAndHashFromPath(path: string): string {
...
+
+  Object.keys(values).forEach((key) => {
+    const value = values[key];
+
+    if (typeof value === "boolean") {
+      ms.replaceAll(key, value ? "true" : "false");
+    }
+  });
+
+  if (ms.hasChanged()) {
```
lforst, line 326: this is nice!

packages/bundler-plugin-core/src/utils.ts
```diff
@@ -307,3 +308,27 @@ export function stripQueryAndHashFromPath(path: string): string {
...
+  code: string,
+  values: Record<string, boolean | undefined>
+): { code: string; map: SourceMap } | null {
+  const ms = new MagicString(code);
+
+  Object.keys(values).forEach((key) => {
+    const value = values[key];
+
+    if (typeof value === "boolean") {
+      ms.replaceAll(key, value ? "true" : "false");
```
lforst, line 322:
```suggestion
      ms.replaceAll(key, JSON.stringify(value));
```

It's a bit more flexible but does the same thing.

packages/bundler-plugin-core/src/index.ts
```diff
@@ -161,6 +164,31 @@ export function sentryUnpluginFactory({
       })
     );
 
+    if (options.bundleSizeOptimizations) {
+      const { bundleSizeOptimizations } = options;
+      const values: SentrySDKBuildFlags = {};
```
lforst, line 169: l: I would maybe name this something like stringReplacementValues or something like that.

packages/webpack-plugin/src/index.ts
```diff
@@ -45,6 +46,31 @@ function webpackReleaseInjectionPlugin(injectionCode: string): UnpluginOptions {
...
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore webpack version compatibility shenanigans
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+      const DefinePlugin =
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore webpack version compatibility shenanigans
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        compiler?.webpack?.DefinePlugin ||
+        webback4or5?.DefinePlugin ||
+        webback4or5?.default?.DefinePlugin;
```
lforst, line 62: beautiful :D

lforst:
> Do you think there's a way that we could filter out files before reading them? I'm thinking about looking at the file path and checking for a @sentry occurance. But not sure if I'm missing something tough why this wouldn't work.

Not a bad idea. we could do `id.contains(os.sep + "node_modules" + os.sep + "@sentry" + os.sep)`. This has a few cases with resolution but in 99% of the cases it should work.

Counterargument, it would not be the same behavior for all bundlers and why should we make this unnecessarily fragile?

Lms24:
> Counterargument, it would not be the same behavior for all bundlers and why should we make this unnecessarily fragile? 

Also a good point. My concern is mostly around the plugin having to read every file might have a performance impact. But I'm totally fine if we just go with this first and optimize later if necessary.

packages/bundler-plugin-core/src/index.ts
mydea: Tbh I just copied this behaviour from the rollup replace Plugin 😅 they to it that way there. But I can just remove the chunk hook!

mydea:
> Do you think there's a way that we could filter out files before reading them? I'm thinking about looking at the file path and checking for a @sentry occurance. But not sure if I'm missing something tough why this wouldn't work.

Yeah we can for sure revisit this, since this is opt in I think we can tweak this more later if needed!

packages/webpack-plugin/src/index.ts
mydea: 🎨

packages/bundler-plugin-core/src/index.ts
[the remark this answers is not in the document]
mydea: I will try removing renderChunk and see how it goes. we can always revisit this and add it again later!

packages/esbuild-plugin/src/index.ts
```diff
@@ -226,11 +233,33 @@ function esbuildDebugIdUploadPlugin(
...
+
+          const contents = replaceBooleanFlagsInCode(source, values);
+
+          if (!contents) {
+            return null;
+          }
+
+          return { contents: contents.code, loader: "default" };
+        });
+      },
```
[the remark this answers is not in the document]
mydea: just to be clear, as I am not an expert at this at all - we do not do this for the other esbuild plugins, which is where I basically copied this from (as well as looking at some esbuild replace plugin). or is this only because we match all files here? E.g. here: https://esbuild.github.io/plugins/#using-plugins they also just return this directly...?

mydea:
OK so I fixed the tests and tried to address the feedback - is this ready to go? :) 

I did not add any file filtering for now, I think we can still add this at a later stage if we find out this is too slow! This does not run at all if no `bundleSizeOptimizations` are configured so it should not affect any existing users.

packages/esbuild-plugin/src/index.ts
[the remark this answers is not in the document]
lforst: In the other plugins, we are doing very complex logic to make sure all of the user files are still loaded via the plugins they defined. There we are using a sort of proxy logic.

The `onLoad` hook is simply the wrong approach for this. The onLoad hook, as it suggests is for "reading" files off a particular location. For example, esbuild doesn't know how to files that lie somewhere on the internet (http) so you can define an onLoad hook that takes care of that for esbuild. A file can only be loaded once. So if our plugin loads it, no other plugin will be able to. Doing this in our case will result in catastrophic behavior for users that use any kind of loader plugin - in the way it is implemented right now this will pretty much result in catastrophic behavior for ALL users since we always define the load hook.

Esbuild doesn't have any sort of `transform` hook and that is by design to keep it fast. It does all the transpilation and stuff by itself. We _need_ to use the `define` option.

lforst:
[This concern](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/428#discussion_r1366770708) is critical and we need to address it before merging this.

packages/esbuild-plugin/src/index.ts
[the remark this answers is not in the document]
mydea: Ah, that makes sense, thank you for the in depth explanation! I'll adjust this then according to this 🙏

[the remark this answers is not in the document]
mydea: I updated this to use `define` (which is much nicer anyhow!)

packages/esbuild-plugin/src/index.ts
```diff
@@ -239,18 +237,14 @@ function esbuildBundleSizeOptimizationsPlugin(
...
-          }
-
-          return { contents: contents.code, loader: "default" };
-        });
+      setup({ initialOptions }) {
+        const replacementStringValues: Record<string, string> = Object.entries(
+          replacementValues
+        ).reduce((acc, [key, value]) => {
+          return { ...acc, [key]: JSON.stringify(value) };
+        }, {});
```
lforst, line 245: this is an accidental O(n^2) because of the spread in a loop, I don't think it matters a lot but since this is an easy fix we should just fix it.

lforst (approved):

Issue #435, opened 2023-11-08
`unplugin` not update to latest version


rchl:
### Environment

`@sentry/bundler-plugin-core` v2.9.0

### Steps to Reproduce

With latest `rollup` installed (3.29.4 now) in my project I'm getting an error like:

```
node_modules/unplugin/dist/index.d.ts(1,40): error TS2305: Module '"rollup"' has no exported member 'AcornNode'.
```

It seems like `unplugin` is pinned here and the latest version of it doesn't reference `AcornNode` anymore so updating should fix this issue at least.

AbhiPrasad:
@lforst does this issue still apply?

lforst:
yup still applies

PR #438, opened 2023-11-09
deps: Bump unplugin to 1.5.0

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/435


mydea (approved):

AbhiPrasad (commented):
Should we also bump in all the individual bundler packages?

lforst:
Totally forgot that it is a dep in all packages.

AbhiPrasad (approved):

lforst:
I cannot be bothered to fix this. I am gonna remove unplugin as a dependency.

Issue #434, opened 2023-11-03
Check for mismatched installed Sentry SDK packages


mydea:
Having multiple different versions of SDK packages installed can lead to a lot of subtle and hard-to-debug issues. This can sadly happen quite easily, e.g. if you have a bundler plugin installed and then install a newer sentry SDK package, you may still have some references to the versions previously installed through the bundler plugin installed, etc. We have this issue pop up quite regularly, and right now it is very hard for either us our our users to figure out what is going on, and if that is a problem.

In order to fix this, we should add three things

* A command to check if you have mismatched SDK versions installed: `npx @sentry/bundler-plugin-core@latest verify-sdk-versions`
* Automatically run this command in bundler builds & output a helpful warning if it fails
* A command to try to _fix_ the mismatched SDK versions: `npx @sentry/bundler-plugin-core@latest fix-sdk-versions`

These commands would need to work for npm, yarn & pnpm lockfiles. I propose to use an approach similar to https://github.com/mydea/yarn-update-dependency, where we actually read the lockfile and find all the installed versions from there. We can probably also use a similar approach to try to _fix_ the installed versions.

PR #439, opened 2023-11-14
feat(core): Warn when mismatching Sentry packages are installed

mydea:
This PR ensures we log out a message when the bundler plugins run that warns you if we detect multiple different versions of the same package being installed in your npm/yarn/pnpm lockfile.

* This parses the lockfiles themselves and checks the installed versions in there directly
* We do not support bun because the lockfile is binary there, so I figure this is a bit more work (we may add this later, if we want)
* We simply log a warning message out, telling you which packages are installed with which versions. For now, users have to manually resolve this.

Open questions:
* For now I consider any package starting with `@sentry/xxx`, _and_ `@sentry-internal/tracing`. Is that a correct assumption? 🤔 Should we ignore `@sentry/cli`, for example? Or any other internal packages to cover as well?

Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/434


lforst:
While it is a good idea, I don't think we'll merge this. Imo it's pushing the scope of the plugins a bit too much.

PR #440, opened 2023-11-14
chore: bump @sentry/cli dependency to 2.21.4

xpirt:
we want this fix: https://github.com/getsentry/sentry-cli/pull/1815

thanks.

### DIFF
--- a/packages/bundler-plugin-core/package.json
+++ b/packages/bundler-plugin-core/package.json
@@ -52,7 +52,7 @@
     "fix": "eslint ./src ./test --format stylish --fix"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mydea:
There is no need to bump this package to get the sentry-cli fix - this is covered by the defined version range, you can just run e.g. `yarn upgrade` or a similar command to update your dependencies to their latest matching in the defined version range!

mydea (approved):

xpirt:
> There is no need to bump this package to get the sentry-cli fix - this is covered by the defined version range, you can just run e.g. `yarn upgrade` or a similar command in your app to update your dependencies to their latest matching in the defined version range!

thanks!

PR #454, opened 2023-12-08
fix: use in memory esbuild build artifact for default source maps

wSedlacek:
This fixes an issue when trying to use the esbuild plugin with Angular (or other similar tool chains with `write` set to false)
This does not modify the behavior when an explicit `globAssets` is set (those will still be limited to what is on disk) but at least for the default option the in memory versions will be used.
This may have a benefit for esbuild since it isn't having to read all these files from disk.

fixes: #351

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -35,6 +35,15 @@ interface DebugIdUploadPluginOptions {
   };
 }
 
+export interface InMemoryBundleAsset {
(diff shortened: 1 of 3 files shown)


wSedlacek:
I have not tested this (hence it being marked as a draft)
I will probably copy over the changes as a patch to our repo and test it there on Monday.

wSedlacek:
I am going to close this because I am not intending to peruse it further at this time.
Anyone who continues this problem in the future is welcome to reference or use the changes here.

PR #459, opened 2024-01-10
fix: Bundle Sentry SDK dependency

lforst:
Ref: https://github.com/getsentry/sentry-javascript/issues/10121
Ref: https://github.com/getsentry/sentry-javascript/issues/9981#issuecomment-1881151710

We should bundle the SDK dependencies so that we do not facilitate issues with multiple different SDK versions in user's apps.

### DIFF
--- a/packages/bundler-plugin-core/package.json
+++ b/packages/bundler-plugin-core/package.json
@@ -53,13 +53,12 @@
   },
   "dependencies": {
     "@sentry/cli": "^2.22.3",
-    "@sentry/node": "^7.60.0",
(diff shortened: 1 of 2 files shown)


mydea (approved):

lforst:
We had to rever this because it caused issues with .mjs rollup configs because of the `https-proxy-agent` module >:[

filpgc:
any update on this?

AbhiPrasad:
We released `7.94.1` with vendoring in `http-proxy-agent`, we can have another go at releasing this now

Issue #461, opened 2024-01-11
Getting intermittent "socket hang up" errors during github actions build process


lsnider:
### Environment

sentry-vite-plugin in github actions

### Steps to Reproduce

During the website build process, we are getting intermittent '[sentry-telemtry-plugin] socket hang up' errors.  It used to happen maybe 1 out of 20 builds.  Recently, it's happening on most builds.  We often have to try the build process 5 to 10 times to get a successful build.

### Expected Result

Not error on building

### Actual Result

I've attached a screenshot of the error during the build process in github actions.

![Screenshot 2024-01-10 at 3 40 50 PM](https://github.com/getsentry/sentry-javascript-bundler-plugins/assets/5577089/6365fbdd-5bf1-45a1-9361-9d13d1fdb3b5)

lforst:
Oof that's not good. I'll take a look at that. In the meanwhile can you try setting `telemetry: false` in the Sentry Vite Plugin options? That should hopefully unblock you.

PR #462, opened 2024-01-12
fix(core): Safely flush telemetry

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/461

I keep forgetting that flushing may throw.


Lms24 (approved):
I wonder if this needs fixing on the SDK side. Should it even throw if a flush fails?

Anyway, we should merge this for now to unblock folks

PR #464, opened 2024-01-22
feat(core): Automatically annotate React components at build time

0Calories:
Adds an **opt-in** feature to the bundler plugins that will automatically add annotations to React components in JSX and TSX files, and their resulting HTML DOM nodes. This is done via an underlying Babel plugin that is run during build time. To enable the feature, set `reactAnnotate.enabled` to `true`.

Example of an automatically annotated component in Sentry's frontend:

image

This will allow the browser SDK to pick off component names and send them to Sentry, allowing for some very useful features:

- React component names in Replays UI
- React component name search / indexing in Replays
- React component names in Breadcrumbs
- React component names in interaction transactions, grouping spans by component name

image

The groundwork of the above features are already in place, now we just need to rollout this plugin to customers and observe! Adding it in the bundler plugins allows for an easy way to gain access to this feature, without requiring our customers to muck about in the build configs and manually add the Babel plugin.

This PR comes with integration tests, and has also been manually tested on all supported bundlers: **Vite, ESBuild, Rollup, Webpack** 

Due to the experimental nature of this feature, it's opt in for now, but we eventually will want it enabled by default so as many customers as possible can get access to the capabilities that it unlocks. Of course, loader logic will need to be added where applicable so that performance is not impacted


packages/bundler-plugin-core/package.json
```diff
@@ -52,6 +52,8 @@
     "fix": "eslint ./src ./test --format stylish --fix"
   },
   "dependencies": {
+    "@babel/core": "7.18.5",
```
0Calories, line 55: Had to move this to a regular dependency, since it is necessary to run `transformAsync` from `@babel/core` during the transform hook

0Calories:
In the process of fixing some lint errors, but the bulk of this PR is done now. Opening it for review

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+      this.excludedComponents = this.opts[excludedComponentsOptionName] || [];
+      if (this.opts.setFSTagName && !this.opts.native) {
+        throw new Error(
+          "`setFSTagName: true` is invalid unless `native: true` is also set in the configuration for @fullstory/babel-plugin-annotate-react"
+        );
+      }
+    },
+    visitor: {
+      FunctionDeclaration(path, state) {
+        if (!path.node.id || !path.node.id.name) return;
```
AbhiPrasad, line 37: l: let's do full return statements here to help with readability

```js
if (!path.node.id || !path.node.id.name) {
  return;
}
```

Ditto with everything else.

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
AbhiPrasad, line 1: Can we make this file TS?

Given the logic is so small, I think we can get away with making any changes we need, and I'd rather do this for future maintainability.

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+ */
+
+const webComponentName = "data-sentry-component";
+const webElementName = "data-sentry-element";
+const webSourceFileName = "data-sentry-source-file";
+
+const nativeComponentName = "dataSentryComponent";
+const nativeElementName = "dataSentryElement";
+const nativeSourceFileName = "dataSentrySourceFile";
+const fsTagName = "fsTagName";
```
AbhiPrasad, line 13: m: do we want to have `fsTagName`?

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
+/**
+ * This plugin is vendored in from FullStory's Babel React Annotate plugin:
+ * https://github.com/fullstorydev/fullstory-babel-plugin-annotate-react/tree/master
```
AbhiPrasad, line 3: h: Vendoring in requires us to copy paste license.

If you make the review changes recommended we should also add notes about what things we changed.

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+}
+
+function sourceFileNameFromState(state) {
+  const name = fullSourceFileNameFromState(state);
+  if (name === undefined) {
+    return undefined;
+  }
+
+  if (name.indexOf("/") !== -1) {
+    return name.split("/").pop();
```
AbhiPrasad, line 110: IMO we should lift up magic constants into it's own variables.

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+  const fullSourceFileName = fullSourceFileNameFromState(state);
+  if (fullSourceFileName == undefined) {
+    return false;
+  }
+
+  for (let i = 0; i < knownIncompatiblePlugins.length; i += 1) {
+    let pluginName = knownIncompatiblePlugins[i];
+    if (
+      fullSourceFileName.includes("/node_modules/" + pluginName + "/") ||
+      fullSourceFileName.includes("\\node_modules\\" + pluginName + "\\")
```
AbhiPrasad, line 128: l: suprised this is not turned on as eslint rule, but we should use template syntax

```suggestion
      fullSourceFileName.includes(`/node_modules/${pluginName}/`) ||
      fullSourceFileName.includes(`\\node_modules\\${pluginName}\\`)
```

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+}
+
+function isKnownIncompatiblePluginFromState(state) {
+  const fullSourceFileName = fullSourceFileNameFromState(state);
+  if (fullSourceFileName == undefined) {
+    return false;
+  }
+
+  for (let i = 0; i < knownIncompatiblePlugins.length; i += 1) {
+    let pluginName = knownIncompatiblePlugins[i];
```
AbhiPrasad, line 125: m: This for loop can be an `array.some` call.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -278,6 +278,17 @@ export interface Options {
     excludeReplayWorker?: boolean;
   };
 
+  /**
+   * Options related to react component name annotations.
+   * Disabled by default, unless a value is set for this option
+   */
+  reactAnnotate?: {
```
AbhiPrasad, line 285: Let's make this option name a bit more detailed

```suggestion
  reactComponentNameAnnotate?: {
```

AbhiPrasad (commented):
Okay initial review - but this PR needs to be split up to make it easier to review/understand.

1. A PR that just vendors in `packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js`, but makes it TS. Ideally we port any unit tests, or write ones as appropriate.
2. A PR that adds `react-annotate-plugin` to core package, adds it to one of webpack/esbuild/vite/rollup, and adds e2e test accordingly
3. A PR that adds plugin to all the rest of the plugins, and updates docs in the package (grep for `OPTIONS_SECTION_INSERT` and `generateOptionsDocumentation()` in the repo for docs info).

packages/bundler-plugin-core/src/index.ts
```diff
@@ -510,6 +532,59 @@ export function createRollupDebugIdUploadHooks(
...
 }
 
+export function createReactAnnotateHooks() {
+  return {
+    async transform(code: string, id: string): Promise<TransformResult> {
+      // id may contain query and hash which will trip up our file extension logic below
+      const idWithoutQueryAndHash = stripQueryAndHashFromPath(id);
+
+      if (idWithoutQueryAndHash.match(/\\node_modules\\|\/node_modules\//)) {
+        return null;
```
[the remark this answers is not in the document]
0Calories: However, we already return `null` in our release injection plugin, under the transform hook for these same situations. Example: https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/main/packages/bundler-plugin-core/src/index.ts#L388

packages/bundler-plugin-core/src/index.ts
```diff
@@ -510,6 +532,59 @@ export function createRollupDebugIdUploadHooks(
...
+
+      try {
+        const result = await transformAsync(code, {
+          plugins: [[reactAnnotate]],
+          filename: id,
+          parserOpts: {
+            sourceType: "module",
+            allowAwaitOutsideFunction: true,
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore This array will always be type compliant when using 'jsx' or 'typescript'
```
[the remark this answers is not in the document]
0Calories:
```
Type 'string[]' is not assignable to type 'PluginConfig[]'.
  Type 'string' is not assignable to type 'PluginConfig'.ts(2322)
(property) ParserOptions.plugins?: PluginConfig[] | undefined
```

I've tried to make this type compliant, but it was a colossal pain. The appropriate type is available only in `@babel/parser`, and I didn't want to import an entire package just for one type. I've tried to use `@babel/types` as well, but it didn't seem like the type I needed was exported from there.

packages/bundler-plugin-core/src/index.ts
```diff
@@ -510,6 +532,59 @@ export function createRollupDebugIdUploadHooks(
...
+      if (idWithoutQueryAndHash.endsWith(".jsx")) {
+        parserPlugins.push("jsx");
+      } else if (idWithoutQueryAndHash.endsWith(".tsx")) {
+        parserPlugins.push("jsx");
+        parserPlugins.push("typescript");
+      }
+
+      try {
+        const result = await transformAsync(code, {
+          plugins: [[reactAnnotate]],
```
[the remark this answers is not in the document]
0Calories: We need to pass in the `filename` to the options which is passed to as an argument to the `transform` hook, so we may not be able to do this because we will still have to reconstruct the object with the constants

packages/bundler-plugin-core/src/plugins/react-annotate-plugin.js
```diff
@@ -0,0 +1,480 @@
...
+  "react-native-testfairy",
+  // This module checks for unexpected property keys and throws an exception.
+  "@react-navigation",
+];
+
+export default function reactAnnotate({ types: t }) {
+  return {
+    pre() {
+      this.excludedComponents = this.opts[excludedComponentsOptionName] || [];
+      if (this.opts.setFSTagName && !this.opts.native) {
```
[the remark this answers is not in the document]
0Calories: Will address this and other comments specific to the plugin code in the PR that will make it its own package

packages/bundler-plugin-core/src/index.ts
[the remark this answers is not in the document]
AbhiPrasad: Can we pick the type from using `Parameters` and `typeof transformAsync`?

packages/esbuild-plugin/src/index.ts
```diff
@@ -41,6 +43,36 @@ function esbuildReleaseInjectionPlugin(injectionCode: string): UnpluginOptions {
...
+          return {
+            loader,
+            pluginName,
+            contents,
+          };
+        });
+      },
+    },
+  };
+}
```
lforst, line 74: Unfortunately, we cannot use a simple `onLoad` hook for this funcitonality esbuild plugin, and in fact, I believe we cannot do this with good conscience in esbuild at all. Esbuild doesn't have a transform hook and that is by design. Read this comment I wrote for a similar situation: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/428#discussion_r1366770708

I don't feel like pressing the "changes requested" button but please consider this as a pretty hard blocker!

0Calories: Hey Luca, thanks for letting me know. I see, that's quite unfortunate that we can't use `onLoad` like this :/ 
Would you know of any alternatives that we can use to support this functionality in esbuild?

I assume the biggest problem here is that the `onLoad` hook only loads in files once, so this will prevent other loaders from running on these same files. In that case, can we fix this by using a `namespace`?

The esbuild docs says this:

> A callback added using onLoad will be run for each unique path/namespace pair that has not been marked as external.

It sounds like as long as we assign a unique namespace during the `onResolve` hook, we can get around this. Please lmk if this is an acceptable solution :)

0Calories: The plugin should not add much computational overhead as well, since it is an O(n) time algorithm. All it's doing is parsing JSX files and adding additional properties onto JSX elements

lforst: I honestly do not think you can do this in esbuild with good conscience. If you assign a different namespace to the files in onResolve you will prevent the file from being picked up by other plugins' onLoad hooks and breaking them. This might be fine in a lot of cases but as soon as people add their own onLoad hooks this will break. I'll leave it up to you but just so you're aware.

0Calories: Understood, in that case we will hold off on releasing this plugin to esbuild for now, but we can reconsider it in the future if customers ask for it 👍

packages/bundler-plugin-core/src/index.ts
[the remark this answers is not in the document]
0Calories: I tried that yesterday but was not successful 😢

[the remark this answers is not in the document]
AbhiPrasad: `NonNullable<NonNullable<Parameters<typeof transformAsync>[1]>['parserOpts']>['plugins']` should work!

0Calories:
Closing this PR, I'm currently in the process of breaking it down into incremental PRs

packages/bundler-plugin-core/src/index.ts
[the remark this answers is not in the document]
0Calories: Nice, that did the trick!

PR #468, opened 2024-01-31
feat(component-annotate): Introduce new plugin to annotate frontend components at build-time

0Calories:
Originally from this PR, splitting it into incremental smaller PRs and also making some adjustments **RE**: Abhi's suggestions:
https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/464

This PR in particular adds a new package to the bundler plugins; a Babel plugin that will annotate frontend components at build-time with additional data. A lot of the source code is vendored from FullStory's React annotate plugin, but has been modified and converted to **TypeScript**. For now, the plugin only works on React, but we can extend it to other frontend frameworks / libraries in the future.

Example of an automatically annotated component in Sentry's frontend:

image

This will allow the browser SDK to pick off component names and send them to Sentry, allowing for some very useful features:

- React component names in Replays UI
- React component name search / indexing in Replays
- React component names in Breadcrumbs
- React component names in interaction transactions, grouping spans by component name

image


packages/component-annotate-plugin/README_TEMPLATE.md
AbhiPrasad, line 1: Let's make this a README with some explanations about how it works?

packages/component-annotate-plugin/test/test-plugin.test.ts
```diff
@@ -0,0 +1,2238 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+import { transform } from "@babel/core";
+import { componentNameAnnotatePlugin as plugin } from "../src/index";
+
+const BananasPizzaAppStandardInput = `import React, { Component } from 'react';
```
AbhiPrasad, line 29: can we make these separate JS files we import and assert on?

You should be able to use `fs.readFileSync` or similar.

packages/component-annotate-plugin/src/index.ts
```diff
@@ -0,0 +1,510 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+/**
+ * The following code is based on the FullStory Babel plugin, but has been modified to work
+ * with Sentry products
```
AbhiPrasad, line 28: let's make sure we detail what specifically changed between vendored code.

packages/component-annotate-plugin/src/constants.ts
```diff
@@ -0,0 +1,121 @@
...
+  // This module might be causing an issue preventing clicks. For safety, we won't run on this module.
+  "react-native-testfairy",
+  // This module checks for unexpected property keys and throws an exception.
+  "@react-navigation",
+];
+
+export const DEFAULT_IGNORED_ELEMENTS = [
+  "a",
+  "abbr",
+  "address",
```
AbhiPrasad, line 11: Are these vendored as well? We need to add license if so.

packages/component-annotate-plugin/src/index.ts
```diff
@@ -0,0 +1,510 @@
...
+const webSourceFileName = "data-sentry-source-file";
+
+const nativeComponentName = "dataSentryComponent";
+const nativeElementName = "dataSentryElement";
+const nativeSourceFileName = "dataSentrySourceFile";
+
+interface AnnotationOpts {
+  native?: boolean;
+  "annotate-fragments"?: boolean;
+  ignoreComponents?: IgnoredComponent[];
```
AbhiPrasad, line 47: Will we expose any of these to users?

If so we should add doc strings and rename `annotate-fragments` to `annotateFragments`

packages/component-annotate-plugin/src/index.ts
```diff
@@ -0,0 +1,510 @@
...
+          t,
+          path,
+          path.node.id.name,
+          sourceFileNameFromState(state),
+          attributeNamesFromState(state),
+          state.opts.ignoreComponents ?? []
+        );
+      },
+      ArrowFunctionExpression(path, state) {
+        const parent = path.parent; // We're expecting a `VariableDeclarator` like `const MyComponent =`
```
AbhiPrasad, line 80: let's move comment above statement here.

packages/component-annotate-plugin/src/index.ts
```diff
@@ -0,0 +1,510 @@
...
+          !parent ||
+          !("id" in parent) ||
+          !parent.id ||
+          !("name" in parent.id) ||
+          !parent.id.name
+        ) {
+          return;
+        }
+
+        if (isKnownIncompatiblePluginFromState(state)) return;
```
AbhiPrasad, line 91:
```suggestion
        if (isKnownIncompatiblePluginFromState(state)) {
          return;
        }
```

ditto with other `return` cases.

0Calories:
Still some tweaks I need to do on this PR, but I've opened it for review for the time being

AbhiPrasad (commented):
Looks good! Thanks for splitting it up :)

packages/component-annotate-plugin/README_TEMPLATE.md
0Calories: Yep, writing that right now!

packages/component-annotate-plugin/src/index.ts
0Calories: Not exposing this for the time being

packages/component-annotate-plugin/test/test-plugin.test.ts
0Calories: I'm going to put this in the backlog for now, eventually when we extend this plugin with more functionality, I want to overhaul the test suite. I'm not entirely satisfied with the vendored tests (coverage is good but tests could be cleaner), but I want to focus on shipping this first

packages/component-annotate-plugin/README_TEMPLATE.md
```diff
@@ -0,0 +1,82 @@
...
+</p>
+
+# Sentry Component Annotate Plugin
+
+[![npm version](https://img.shields.io/npm/v/@sentry/component-annotate-plugin.svg)](https://www.npmjs.com/package/@sentry/component-annotate-plugin)
+[![npm dm](https://img.shields.io/npm/dm/@sentry/component-annotate-plugin.svg)](https://www.npmjs.com/package/@sentry/component-annotate-plugin)
+[![npm dt](https://img.shields.io/npm/dt/@sentry/component-annotate-plugin.svg)](https://www.npmjs.com/package/@component-annotate-plugin)
+
+A Babel plugin that automatically annotates your output DOM with their respective frontend component names.
+This will unlock the capability to search for Replays in Sentry by component name, as well as see component names in breadcrumbs and performance monitoring.
```
AbhiPrasad, line 14: Let's add minimum SDK version needed, and rename this file to be `README.md`

Also can we add a note that this is in a beta state? See https://github.com/getsentry/sentry-javascript/blob/develop/packages/bun/README.md for an example of a quick beta note.

packages/component-annotate-plugin/package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "@sentry/component-annotate-plugin",
+  "version": "1.0.0",
```
AbhiPrasad, line 3: This needs to align with the other version of the other packages in this repo.

0Calories:
Addressed your comments @AbhiPrasad

AbhiPrasad (approved):

PR #469, opened 2024-01-31
feat(core): Include component name annotation plugin with all bundler plugins except esbuild

0Calories:
Breaking up this PR into a smaller chunk: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/464/

This PR focuses on bringing in the new plugin introduced in https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/468, and adding it to all bundlers as an optional feature. It is not currently supported by esbuild, so there are guardrails preventing its use in the esbuild bundler plugins.


0Calories:
Still need to do some finishing work on this PR but opening it early to get some eyes on it due to the timezone difference in VIE. In the morning, I will update the documentation in the README to include these changes

packages/bundler-plugin-core/src/index.ts
```diff
@@ -317,6 +323,22 @@ export function sentryUnpluginFactory({
       );
     }
 
+    if (!options.componentNameAnnotate) {
+      logger.warn(
+        "No options provided for the component name annotate plugin. Please set `componentNameAnnotatePlugin.enabled` to `true` if you would like to have your frontend components automatically annotated at build-time."
+      );
+    } else if (!options.componentNameAnnotate.enabled) {
```
lforst, line 330: We should not print this log message generically. It wouldn't make any sense in non-react applications.

packages/bundler-plugin-core/src/types.ts
```diff
@@ -278,6 +278,21 @@ export interface Options {
     excludeReplayWorker?: boolean;
   };
 
+  /**
+   * Options related to component name annotations.
+   * Disabled by default, unless a value is set for this option.
+   * Please note that this feature is not currently supported by the esbuild bundler plugins, and will only annotate React components
+   */
+  componentNameAnnotate?: {
```
lforst, line 286: The name of this option should properly reflect that it will only annotate components in react applications. Probably something like `reactComponentAnnotation`.

I would also hoist the JS doc for the enabled option to this level and only explain what enabled does in the enabled js doc.

lforst (approved):
Generally looks good! 👍 I left some comments that should be adressed.

Please also add documentation for the option to `generate-documentation-table.ts` so that the readmes on npm have the the option. It would be extra cool if you could somehow manage to exclude the added option from the docs (and also the types) for esbuild.

packages/bundler-plugin-core/src/types.ts
0Calories: I kept the name generic for now since we eventually may extend it to work for other frameworks / libraries, and didn't want to have users be forced to change the option in their config when that does happen.

Noted on the docs, will change that!

packages/bundler-plugin-core/src/index.ts
0Calories: True, I will remove it in that case then

packages/bundler-plugin-core/src/types.ts
0Calories: alright I'll change it 👍

0Calories:
@lforst Addressed all your comments and added conditional generation of options in the docs based on the bundler type 😄

PR #474, opened 2024-02-02
ref(component-annotate): Prefix plugin name with `babel`

0Calories:
The Babel loader is unable to resolve plugins when adding it in the babel config file via their package name string, if their package name is not prefixed with `babel`.

Example: [This does not work](https://github.com/getsentry/sentry/pull/64439/files#diff-e5806ce5b4f1576f761dd99bdfd977d227f0c514d67dc3103d042f67784a66f9R54), the Babel loader fails to resolve the plugin because it automatically prefixes it with `babel`


packages/component-annotate-plugin/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "@sentry/component-annotate-plugin",
+  "name": "@sentry/babel-component-annotate-plugin",
   "version": "2.11.0",
   "description": "A Babel plugin that annotates frontend components with additional data to enrich the experience in Sentry",
   "repository": "git://github.com/getsentry/sentry-javascript-bundler-plugins.git",
-  "homepage": "https://github.com/getsentry/sentry-javascript-bundler-plugins/tree/main/packages/component-annotate-plugin",
+  "homepage": "https://github.com/getsentry/sentry-javascript-bundler-plugins/tree/main/packages/babel-component-annotate-plugin",
```
AbhiPrasad, line 6: We should rename the folder then right?

0Calories: oh right

AbhiPrasad (approved):

PR #478, opened 2024-02-05
ref(component-annotate): Use default export

0Calories:
I know, default exports bad, but we don't have much of a choice here. Babel expects our plugin to be using a default export, or else it will not be able to resolve the plugin when using its string identifier in the config (`@sentry/babel-plugin-component-annotate`)

See this in Babel's source for reference:
https://github.com/babel/babel/blob/main/packages/babel-core/src/config/config-descriptors.ts#L335

### DIFF
--- a/packages/babel-plugin-component-annotate/README.md
+++ b/packages/babel-plugin-component-annotate/README.md
@@ -69,7 +69,7 @@ Or alternatively, configure the plugin by directly importing it:
 ```js
 // babel.config.js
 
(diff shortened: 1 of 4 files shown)


AbhiPrasad (approved):
Thanks babel.

Can we add a comment about why it has to be default export?

lforst (approved):

0Calories:
Good idea, added a comment 👍

Issue #484, opened 2024-02-13
@sentry/vite-plugin inclusion in vite.config.js causes it to skip .env.production


719media:
### Environment

What version are you running? Etc.

### Steps to Reproduce

1. npm install @sentry/vite-plugin
2. .env.production files are no longer used. Major problem! (however, .env.production.local files appear to still be used)

### Expected Result

.env.production files should be used

### Actual Result

.env.production files are no longer used, and instead just .env (and .env.local, .env.[name].local) are brought in

719media:
Sorry, I know the report is sparse, I will dig a little more, but I'm going to guess it has something to do with
[sentry-vite-plugin] Info: Using environment variables configured in ".env.sentry-build-plugin".

lforst:
Hi, I tried this with the steps you outlined but couldn't reproduce the issue you're describing. Would you mind creating a reproduction repo we could pull to replicate this? Thank you!

719media:
Sure. Here you go
https://github.com/719media/sentry-vite-env-bug

lforst:
Awesome thank you. Found the bug. Will release it asap.

PR #486, opened 2024-02-15
fix(core): Stop .env files from being picked up

lforst:
For some reason `dotenv` is falling back to reading the local `.env` file when the file you actually wanted to read is not present. Obviously this is not what we want so instead we just read it ourselves and just use dotenv to parse it for us.

Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/484

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -83,7 +83,7 @@ export function sentryUnpluginFactory({
         path.join(process.cwd(), ".env.sentry-build-plugin"),
         "utf-8"
       );
-      // NOTE: Do not use the dotenv library directly to read the dotenv file! For some ungodly reason, it falls back to reading `${process.cwd()}/.env` which is absolutely not what we want.
+      // NOTE: Do not use the dotenv.config API directly to read the dotenv file! For some ungodly reason, it falls back to reading `${process.cwd()}/.env` which is absolutely not what we want.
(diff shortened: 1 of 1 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -78,17 +78,20 @@ export function sentryUnpluginFactory({
...
-    if (dotenvResult.error && dotenvResult.error.code !== "ENOENT") {
-      throw dotenvResult.error;
-    } else if (dotenvResult.parsed) {
+    try {
+      const dotenvFile = fs.readFileSync(
+        path.join(process.cwd(), ".env.sentry-build-plugin"),
+        "utf-8"
+      );
+      // NOTE: Do not use the dotenv library directly to read the dotenv file! For some ungodly reason, it falls back to reading `${process.cwd()}/.env` which is absolutely not what we want.
+      const dotenvResult = dotenv.parse(dotenvFile);
```
Lms24, line 87: I guess this is what we actually shouldn't use, right? Given we still call `dotenv.parse` below
```suggestion
      // NOTE: Do not use the dotenv.config API directly to read the dotenv file! For some ungodly reason, it falls back to reading `${process.cwd()}/.env` which is absolutely not what we want.
      const dotenvResult = dotenv.parse(dotenvFile);
```

Lms24 (approved):

Issue #501, opened 2024-03-11
Cannot read properties of undefined (reading "silent")


Tofandel:
Error is happening on this line

https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/fe6169b73a8f5a95d9ac2c263ddb592756ec90f4/packages/bundler-plugin-core/src/index.ts#L76

### Environment

When calling `sentryVitePlugin()` without any configuration option

lforst:
Hi, we just released a "fix" for this in the latest release!

PR #502, opened 2024-03-11
feat: Make options object optional

lforst:
Resolves https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/501

It's actually very valid as the plugin doesn't need any options to operate. Everything necessary for source maps upload can also be configured via environment variables.


Lms24 (approved):
Good point!

PR #517, opened 2024-04-09
feat: Support Heroku env vars when infering release name

duailibe:
This PR checks Heroku env vars when trying to determine a release name.

This should align with other Sentry tools (including the Rust cli) and this library also used to support them a very long time ago:

https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/484a719a187fdfdf456b9da1489b8eed1e7dde19/packages/unplugin/src/getReleaseName.ts#L18-L25

Feel free to not add this if this is not wanted!


lforst (approved):
Great improvement. Thanks!

Issue #481, opened 2024-02-06
Setting `cleanArtifacts: true` fails with a 403


Wedvich:
### Environment

```
@sentry/webpack-plugin@2.10.3
```

### Steps to Reproduce

I added `cleanArtifacts: true` to my config, and tried to overwrite the same release name.

```ts
release: {
  cleanArtifacts: true,
  name: sentryReleaseId,
  setCommits: {
    auto: true,
  },
},
```

### Expected Result

The existing artifact bundle should be deleted, and the new artifacts should be uploaded. When Webpack is done I expect to only see one artifact bundle under the release.

### Actual Result

The upload with the new artifact bundle succeeds, but the old artifacts aren't deleted, and I get the following error in the logs:

```
  DEBUG   2024-02-06 11:53:05.136196 +01:00 > DELETE /api/0/projects/<redacted>/<redacted>/files/source-maps/?name=local-tmp HTTP/1.1                                       
  DEBUG   2024-02-06 11:53:05.136216 +01:00 > Host: sentry.io                                                                                                                
  DEBUG   2024-02-06 11:53:05.136221 +01:00 > Accept: */*                                                                                                                    
  DEBUG   2024-02-06 11:53:05.136225 +01:00 > Connection: TE                                                                                                                 
  DEBUG   2024-02-06 11:53:05.136228 +01:00 > TE: gzip                                                                                                                       
  DEBUG   2024-02-06 11:53:05.136231 +01:00 > User-Agent: sentry-cli/2.26.0 webpack-plugin/2.10.3                                                                            
  DEBUG   2024-02-06 11:53:05.136975 +01:00 > Authorization: Bearer sntrys_e***                                                                                              
  DEBUG   2024-02-06 11:53:05.292127 +01:00 < HTTP/1.1 403 Forbidden                                                                                                         
  DEBUG   2024-02-06 11:53:05.292146 +01:00 < server: nginx                                                                                                                  
  DEBUG   2024-02-06 11:53:05.292150 +01:00 < date: Tue, 06 Feb 2024 10:53:05 GMT                                                                                            
  DEBUG   2024-02-06 11:53:05.292153 +01:00 < content-type: application/json                                                                                                 
  DEBUG   2024-02-06 11:53:05.292156 +01:00 < Content-Length: 63                                                                                                             
  DEBUG   2024-02-06 11:53:05.292159 +01:00 < allow: GET, DELETE, HEAD, OPTIONS                                                                                              
[...]

PR #525, opened 2024-04-17
feat: Deprecate and noop `cleanArtifacts`

lforst:
`cleanArtifacts` was historically needed to remove artifacts before uploading to avoid collisions. Since Sentry can nowadays accept multiple artifacts with the same name per release and prefer the most recent one this option is unnecessary.

Additionally, it causes problems like these: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/481  with org auth tokens so moving forward it should noop.


Lms24 (approved):
yeah let's just noop here, sounds reasonable!

Issue #445, opened 2023-11-17
Cannot find package '_sentry-debug-id-injection-stub'


jansedlon:
### Environment

MacOS
NodeJS with Remix framework and esbuild bundler
"@sentry/esbuild-plugin": "^2.10.1"

### Steps to Reproduce

1. Created Remix project with custom express server
2. Build server using this package.json script `"build:server": "tsx ./other/build-server.ts",`

Content of `./other/build-server.ts`

```ts
import { sentryEsbuildPlugin } from "@sentry/esbuild-plugin";
import esbuild from "esbuild";
import fsExtra from "fs-extra";
import { globSync } from "glob";
import path from "path";
import { fileURLToPath } from "url";

const pkg = fsExtra.readJsonSync(path.join(process.cwd(), "package.json"));

const __dirname = path.dirname(fileURLToPath(import.meta.url));
const here = (...s: Array<string>) => path.join(__dirname, ...s);
const globsafe = (s: string) => s.replace(/\\/g, "/");

const allFiles = globSync(globsafe(here("../server/**/*.*")), {
  ignore: [
    "**/tsconfig.json",
    "**/eslint*",
    "**/__tests__/**",
    "server/dev-server.js",
  ],
});

const entries = [];
for (const file of allFiles) {
  if (/\.(ts|js|tsx|jsx)$/.test(file)) {
    entries.push(file);
  } else {
    const dest = file.replace(here("../server"), here("../server-build"));
    fsExtra.ensureDirSync(path.parse(dest).dir);
    fsExtra.copySync(file, dest);
    console.log(`copied: ${file.replace(`${here("../server")}/`, "")}`);
  }
}

console.log("building...");

esbuild
  .build({
    entryPoints: entries,
    outdir: here("../server-build"),
    target: [`node${pkg.engines.node}`],
    platform: "node",
    format: "esm",
    sourcemap: true,
    plugins: [sentryEsbuildPlugin({
      org: process.env.SENTRY_ORG,
      project: process.env.SENTRY_PROJECT,
      authToken: process.env.SENTRY_AUTH_TOKEN,
      sourcemaps: {
        assets: here("../server-build"),
      },
    })],
    logLevel: "info",
  })
  .catch((error: unknown) => {
    console.error(error);
    process.exit(1);
  });
```

Run server

### Expected Result

It should start up the server.

### Actual Result

It crashes with
```
[...]

PR #526, opened 2024-04-17
misc(esbuild): Log warning when attempting to inject debug IDs with esbuild `bundle` option active

lforst:
Relates to: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/445#issuecomment-2061371373

### DIFF
--- a/packages/esbuild-plugin/src/index.ts
+++ b/packages/esbuild-plugin/src/index.ts
@@ -4,6 +4,7 @@ import {
   getDebugIdSnippet,
   SentrySDKBuildFlags,
 } from "@sentry/bundler-plugin-core";
+import type { Logger } from "@sentry/bundler-plugin-core";
 import type { UnpluginOptions } from "unplugin";
 import * as path from "path";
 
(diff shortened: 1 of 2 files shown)


packages/esbuild-plugin/src/index.ts
```diff
@@ -50,6 +51,12 @@ function esbuildDebugIdInjectionPlugin(): UnpluginOptions {
 
     esbuild: {
       setup({ initialOptions, onLoad, onResolve }) {
+        if (initialOptions.bundle) {
+          logger.warn(
+            "Esbuild's `bundle` option is currently not supported! Esbuild will probably crash now."
```
Lms24, line 56: l: Did I understand the issue correctly that this only happens for `bundle: true`? In this case, maybe we tell users this specifically:

```suggestion
            "Esbuild's `bundle: true` option is currently not supported! Esbuild will probably crash now."
```

Also, is there anything we can recommend them here instead of just saying that their build will crash?

lforst: Absolutely valid point. I updated the wording!

lforst:
@Lms24 would you mind taking another look?

Lms24 (approved):
Whoops, sorry this fell through. LGTM and thx for making the change!

pckilgore:
image

image


I'm confused here cc @lforst

lforst:
omg I think I completely flipped what I meant to do in that PR. Gonna fix it on monday.

Issue #531, opened 2024-05-15
`sourcemaps.rewriteSources` doesn’t apply new source value


niksy:
### Environment

* Sentry Vite plugin, 2.16.1
* Self hosted Sentry, 22.8.0

### Steps to Reproduce

1. Use Sentry Vite plugin
2. Add option `sourcemaps.rewriteSources: (source) => './' + source.replaceAll('../', '')`

### Expected Result

Source from source map should be translated to value defined in `rewriteSources` hook, e.g. `../` should be stripped and `./` should be prepended to source path.

### Actual Result

Default value is used which is relative path from `process.cwd()`, but I’m not sure if that’s valid - I’ve tried to change code directly inside default hook but changes are not applied.

I’m also using `RewriteFrames` integration, but I haven’t seen any changes if it’s removed.

Is this maybe related to self-hosted Sentry? For upload to work I had to use `release.uploadLegacySourcemaps` option so maybe these changes don’t get applied?

Lms24:
Hey @niksy thanks for writing in!

To debug this, would you mind sharing your vite plugin config? 

It could be well-related to self-hosted and `uploadLegacySourcemaps`. Which makes this extra hard to debug 😅 
are you able to update to the latest self-hosted version of Sentry? If so, I think the simpler, debug-id based source map upload apporach should work (tagging @lforst to confirm).

niksy:
> To debug this, would you mind sharing your vite plugin config?

Sure, no problem, here it is:

```js
sentryVitePlugin({
  url: "",
  org: "",
  project: "",
  authToken: "",
  sourcemaps: {
    rewriteSources: (source) => "./" + source.replace("../", ""),
  },
  release: {
    name: commitSha,
    uploadLegacySourcemaps: path.resolve(cwd, "ssr-dist"),
  },
});
```

> are you able to update to the latest self-hosted version of Sentry?

I’ll see what I can do, it would be great if this is fixed just with version bump, it looks like our self-hosted instance version is almost 2 years old 😄

niksy:
Unfortunately, we’re currently unable to easily upgrade our Sentry instance 😞

lforst:
One question: Did you look at the `sources` field of the uploaded artifact (in Sentry) or the local file? `sourcemaps.rewriteSources` only affects the uploaded artifacts.

lforst:
@niksy okay thank you! I will take a look.

lforst:
[...]

PR #535, opened 2024-05-23
docs: Clarify interaction between `uploadLegacySourcemaps` and other options

lforst:
ref https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/531#issuecomment-2126830977


Lms24 (approved):
Good comment!

PR #538, opened 2024-05-31
feat: Promote experimental `moduleMetadata` option to stable

lforst:
Promotes the `moduleMetadata` integration to stable.

### DIFF
--- a/packages/bundler-plugin-core/src/sentry/telemetry.ts
+++ b/packages/bundler-plugin-core/src/sentry/telemetry.ts
@@ -70,7 +70,7 @@ export function setTelemetryDataOnHub(options: NormalizedOptions, hub: Hub, bund
     );
   }
 
-  hub.setTag("module-metadata", !!options._experiments.moduleMetadata || !!options.moduleMetadata);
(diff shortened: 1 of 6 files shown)


packages/bundler-plugin-core/src/sentry/telemetry.ts
```diff
@@ -70,7 +70,7 @@ export function setTelemetryDataOnHub(options: NormalizedOptions, hub: Hub, bund
     );
   }
 
-  hub.setTag("module-metadata", !!options._experiments.moduleMetadata);
+  hub.setTag("module-metadata", !!options._experiments.moduleMetadata || !!options.moduleMetadata);
```
Lms24, line 73: Are we setting this tag before we normalize options? If not, we could remove the _experimental flag, right?

Lms24 (approved):

packages/bundler-plugin-core/src/sentry/telemetry.ts
lforst: oh you're right. This is after normalization. 👌

PR #539, opened 2024-05-31
feat: Allow passing of meta-framework as telemetry data

lforst:
Allows us to gauge usage across meta frameworks like Next.js, Sveltekit, Solidstart, or Nuxt.


Lms24 (approved):
Nice!

PR #540, opened 2024-05-31
feat: Add `applicationKey` option to identify application code from within the SDK

lforst:
Ties into https://github.com/getsentry/sentry-javascript/pull/12267
Ref https://github.com/getsentry/sentry-javascript/issues/10882

Allows users to mark their bundles with an application key which will allow the SDK to filter events based on application keys in stack frames.

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -224,8 +224,18 @@ export function sentryUnpluginFactory({
       plugins.push(releaseInjectionPlugin(injectionCode));
     }
 
-    if (options.moduleMetadata) {
-      let metadata: Record<string, unknown>;
+    if (options.moduleMetadata || options.applicationKey) {
+      let metadata: Record<string, unknown> = {};
+
+      if (options.applicationKey) {
+        // We use different keys so that if code receives multiple bundling passes.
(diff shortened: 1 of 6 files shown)


timfish (approved):

packages/bundler-plugin-core/src/index.ts
```diff
@@ -224,19 +224,29 @@ export function sentryUnpluginFactory({
...
+    if (options.moduleMetadata || options.applicationKey) {
+      let metadata: Record<string, unknown> = {};
+
+      if (options.applicationKey) {
+        // We use different keys so that if code receives multiple bundling passes.
+        // It is a bit unfortunate that we have to inject the metadata snippet at the top, which means that if we didn't
+        // have this mechanism, the first bundling pass would always "win". If this weren't the case we would be fine with
+        // the last pass winning but the first pass winning is very bad, because it would prevent any sort of overriding.
+        // We can simply use the `_sentryBundlerPluginAppKey:` to filter for app keys in the SDK.
+        metadata[`_sentryBundlerPluginAppKey:${options.applicationKey}`] = true;
```
lforst, line 236: I know this is sus but I hope the comment makes sense

Lms24 (approved):

Issue #537, opened 2024-05-31
Esbuild Crash Warning


jd-carroll:
During build I receive warnings like:
```
[sentry-esbuild-plugin] Warning: Esbuild's `bundle: ***` option is currently not supported! Esbuild will probably crash now. Sorry about that. If you need to upload sourcemaps with the `bundle` option, it is recommended to use Sentry CLI instead: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/
```

However, esbuild never crashes and sourcemaps appear to be uploaded correctly?

Is this actually an issue? If so, is there a summary of the issue somewhere so that I can understand if I will potentially run into these crashes?

lforst:
Interesting. We print this warning as a response to this issue https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/445. Is everything working for you, even though you have `bundle: true` being set. If  so, what esbuild version are you using?

jd-carroll:
I am using `0.20.4`

But I am a little confused, is the issue encountered when `bundle: true` or `bundle: false`?

Because the original description of #445 would be `bundle: false`, however #562 talks about it being an issue when `bundle: true`. Further [the code](https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/45c7f1c816c7fb5fbff86a8516e22341cb4dd8ce/packages/esbuild-plugin/src/index.ts#L54) is also on a `bundle: true` scenario.

I can see this potentially being an issue for `bundle: false`, but I would expect `bundle: true` to work without issue... ?

Lastly, I can see the bundled debugId when I inspect the sources:
```javascript
(t._sentryDebugIds=t._sentryDebugIds||{},t._sentryDebugIds[e]="748e0dce-645a-4821-a7a8-5914b144f971",t._sentryDebugIdIdentifier="sentry-dbid-748e0dce-645a-4821-a7a8-5914b144f971")
```

lforst:
Sorry for not responding here. As you already saw I fumbled the logic 😄 The fix is released now if you upgrade to the latest version!

PR #542, opened 2024-06-03
fix(esbuild): Invert warning about `bundle: true`

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/537

In https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/526 we introduced a log message that esbuild's `bundle: true` is not supported. We messed up because it is actually the complete opposite. We don't support `bundle: false`.

This PR flips the logic.

### DIFF
--- a/packages/esbuild-plugin/src/index.ts
+++ b/packages/esbuild-plugin/src/index.ts
@@ -53,7 +53,7 @@ function esbuildDebugIdInjectionPlugin(logger: Logger): UnpluginOptions {
       setup({ initialOptions, onLoad, onResolve }) {
         if (!initialOptions.bundle) {
           logger.warn(
-            "The Sentry esbuild plugin only supports esbuild with `bundle: true` being set in the esbuild build options! Esbuild will probably crash now. Sorry about that. If you need to upload sourcemaps without `bundle: true`, it is recommended to use Sentry CLI instead: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/"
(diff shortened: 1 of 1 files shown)


andreiborza (approved):

packages/esbuild-plugin/src/index.ts
```diff
@@ -51,9 +51,9 @@ function esbuildDebugIdInjectionPlugin(logger: Logger): UnpluginOptions {
 
     esbuild: {
       setup({ initialOptions, onLoad, onResolve }) {
-        if (initialOptions.bundle) {
+        if (!initialOptions.bundle) {
           logger.warn(
-            "Esbuild's `bundle: true` option is currently not supported! Esbuild will probably crash now. Sorry about that. If you need to upload sourcemaps with the `bundle` option, it is recommended to use Sentry CLI instead: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/"
+            "The Sentry esbuild plugin only supports esbuild with `bundle: true` being set in the esbuild build options! Esbuild will probably crash now. Sorry about that. If you need to upload sourcemaps without `bundle: true`, it is recommended to use Sentry CLI instead: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/"
```
mydea, line 56:
```suggestion
            "The Sentry esbuild plugin only supports esbuild with `bundle: true` being set in the esbuild build options. Esbuild will probably crash now. Sorry about that. If you need to upload sourcemaps without `bundle: true`, it is recommended to use Sentry CLI instead: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/"
```

I would avoid `!`, generally :)

mydea (approved):

jd-carroll:
Awesome! ♥️

PR #547, opened 2024-06-11
fix: Always delete files when `sourcemaps.filesToDeleteAfterUpload` is set

lforst:
Currently we only delete files when the debug ID upload was initiated. This is extremely confusing if you use the legacy upload or didn't upload at all - for example when not providing an auth token.

With this we will always run deletion when `sourcemaps.filesToDeleteAfterUpload` is set, and up to two additional times, after the debug ID upload and the legacy upload have completed.


lforst:
Screenshot 2024-06-17 at 17 23 47

fml

AbhiPrasad (approved):

PR #568, opened 2024-07-10
fix: Use `sequential` and `post` order for vite artifact deletion

tyouzu1:
Sometimes writeBundle is executed before the upload task,  need to ensure the order of code execution


tyouzu1:
No other plugins are considered, currently only for reference

lforst (approved):
Lgtm! Thank you!

tyouzu1:
> Lgtm! Thank you!

Too fast , Hope there are no bugs 🥲

By the way, it seems that my electron vite project(used [vite-plugin-electron](https://github.com/electron-vite/vite-plugin-electron)) has the same problem( Triggered multiple `deletion task` ), Looks like it's all solved.

Thanks again

PR #581, opened 2024-08-05
deps: Bump `@sentry/cli` to `2.33.1`

lforst:


lforst:
@szokeasaurusrex anything we should look out for when bumping the CLI?

szokeasaurusrex:
I cannot think of anything specific @lforst. Maybe you can call out some of the important changes since then:

  - (Soft) validation of auth tokens (warn when invalid format recognized)
  - CLI should now be fully compatible with hybrid cloud EU region
  - A security fix (getsentry/sentry-cli#2115 and getsentry/sentry-cli#2118) to prevent auth tokens from being logged to console in certain situations. This bug might not be applicable to the bundler plugins, though. The tokens only got logged when both of the following were true:
     - The token was passed in via the `--auth-token` argument, not via an environment variable or configuration file, and
     - the CLI's log level was set to `info` or `debug`.

szokeasaurusrex (approved):

PR #582, opened 2024-08-06
feat: Add `bundleSizeOptimizations.excludeTracing` option as alias to deprecated `bundleSizeOptimizations.excludePerformanceMonitoring`

lforst:
Adds an alias because we are rephrasing everything around performance and tracing.

Also adds docs to readme we should have added before.


s1gr1d (approved):
Nice, thanks 🙌🏻

Issue #520, opened 2024-04-11
Webpack plugin uploads source maps on development mode


InterstellarStella:
### Environment

Using `sentry/browser` version `7.109.0` and `sentry/webpack-plugin` version `2.16.1`.

### Steps to Reproduce

1. Create a very minimal JS app and add webpack to it
2. Set up Sentry's SDK and the sentry webpack plugin according to the docs
3. Run the app with `webpack --watch --mode development` OR set `mode:“development”,` in webpack's `module.exports`.

### Expected Result

Source maps are not uploaded to Sentry, as stated [in the docs](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/webpack/):

> The Sentry webpack plugin doesn't upload source maps in watch-mode/development-mode. We recommend running a production build to test your configuration.

### Actual Result

Source maps are uploaded to Sentry.

lforst:
Backlogging this!

lforst:
As a workaround you can set `disable: true` for when you do dev builds.

playaz87:
> As a workaround you can set `disable: true` for when you do dev builds.

This does disable the upload. However it still causes sourcemaps to be generated significantly slowing down the build time.

rabbiveesh:
This is still very broken - even setting `sourcemaps.disable` to an explicit `true` still results in the sourcemaps being uploaded in dev mode.

andreiborza:
As a workaround, setting `sourcemaps.disable` to `true` should now work (see https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/587). 

Backlogging the dev mode improvement.

PR #586, opened 2024-08-12
fix: disable sourcemaps from the modern way of uploading

rabbiveesh:
should fix at least the really nasty part of #520 where it uploads sourcemaps even though
i explicitly configured it not to.

I assume the fix for dev mode would be similar, though I didn't dig into how it tends to
be detected in this project (if at all?)


lforst:
Sorry, didn't see this PR. I'll close in favor of https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/587. I originally didn't mean for the option to affect the legacy upload but the debug Id upload.

PR #589, opened 2024-08-14
fix: Disable debug ID injection when `sourcemaps.disable` is set

lforst:
I ran into the case in https://github.com/getsentry/sentry/pull/76051 where it would be super harmful when we injected debug IDs because we later inject debug IDs with the CLI, and the plugin's debug IDs would win, and source maps would break. This is more or less just a reliability patch.


Lms24 (approved):
makes sense to me!

PR #604, opened 2024-09-10
feat: Update Sentry telemetry to v8

### LINKED ISSUE #579: Update `@sentry/node` to v8
We still use v7 of the Node SDK, we should bump our own usage to v8.

We can also explore using `ServerRuntimeClient` from `@sentry/core` instead of `NodeClient`. We only manually create a client, so maybe we can skip the OTEL dependencies etc. here.

timfish:
Closes #579

Since we bundle the SDK into `bundler-plugin-core`, this PR migrates to using `ServerRuntimeClient` rather than the Node client because we don't need otel instrumentations and it reduces the bundle size by ~300kB. I copied and simplified the Node http transport because referencing the Node SDK at all caused bundling issues.

I also added a way to integration test the telemetry by setting a global which is used to store envelopes rather than sending them.


lforst (approved):
nice, thank you!

Issue #611, opened 2024-09-23
Nextjs builds fail with Sentry and headlessui


brandinchiu:
### Environment

What version are you running? Etc.
"@headlessui/react": "^2.1.5",
"@heroicons/react": "^2.1.5",
"@sentry/nextjs": "^8.28.0",
"next": "14.2.8",
"react": "^18",
"react-dom": "^18"

I've created a point-in-time backup of my current working project here: https://github.com/brandinchiu/sentry-build-error

Disabling `reactComponentAnnotation` will allow the build to complete successfully.

### Steps to Reproduce

1. followed this tutorial to get Sentry installed on a blank Nextjs project: https://docs.sentry.io/platforms/javascript/guides/nextjs/
2. create a page in Nextjs that makes use of a <Menu /> component from headlessui/react
3. attempt to build the application with `reactComponentAnnotation` enabled.

### Expected Result

Build should complete.

### Actual Result

Build fails.  Error is throw indicating that Nextjs prerendering failed when trying to inject Sentry annotations into components:
```
Error occurred prerendering page "/". Read more: https://nextjs.org/docs/messages/prerender-error

Error: Passing props on "Fragment"!

The current component <Menu /> is rendering a "Fragment".
However we need to passthrough the following props:
  - data-sentry-element
  - data-sentry-component
  - data-sentry-source-file
  - data-headlessui-state

You can apply a few solutions:
  - Add an `as="..."` prop, to ensure that we render an actual element instead of a "Fragment".
  - Render a single element as the child so that we can forward the props onto that element.
...
```

chargome:
Hi @brandinchiu, have you tried any of the solutions of not using a fragment?

brandinchiu:
It's an external UI kit, so I have limited control in this case.

Regardless, fragments are a fairly common react construct. Causing all builds to fail if one exists anywhere in the component tree is a bit extreme, imo.

chargome:
@brandinchiu 
Yes, but the error is not originating from Sentry but from headless ui. You could just render the headless menu as a div for example: `<Headless.Menu as="div" {...props} />;`

This should fix your error 👍

brandinchiu:
Respectfully, I disagree with the premise that this is a headlessui issue. I don't even think it's a Sentry issue.

But the Venn diagram here creates a legitimate problem, and I just can't see the solution being "never use a naked fragment as long as Sentry is installed".

[...]

PR #617, opened 2024-10-08
feat(react-component-annotate): Allow skipping annotations on specified components

0Calories:
**Note:** The large number of deleted lines is due to a lot of old tests no longer being relevant since the component ignore config system is being overhauled

Some third party libraries involve components that have compatibility issues with the `react-component-annotate` plugin, and results in errors when the `data-sentry` properties are added as props. This PR introduces a bit of a bandaid fix by introducing an option to allow users to configure the names of components that should **not** have any annotations applied.

When using the standalone Babel plugin, you can configure this option like so:

```
// babel.config.js

{
  // ... other config above ...

  plugins: [
    // Put this plugin before any other plugins you have that transform JSX code
    // The options are set by providing an object as the second element in the array, but not required
    ['@sentry/babel-plugin-component-annotate', {ignoreComponents: ['Foo', 'Bar']}]
  ],
}
```

When using the Sentry bundler plugins, you can configure this option like so:

```
sentryVitePlugin({
    // ... other options
    reactComponentAnnotation: {enabled: true, ignoreComponents: ['Foo', 'Bar']},
}),
```

TODO in future: Allow providing a source file location, since it is possible for multiple components to have the same name, but you may not want to skip annotation on all of these.


lforst:
@0Calories can we expedite this PR?

0Calories:
@lforst I'll try to finish it on friday, I believe it just needs tests

Also:
>  allow providing a source file location, since it is possible for multiple components to have the same name, but you may not want to skip annotation on all of these.

This unfortunately will have to be done in a followup PR. For now, I want to at least expose the option to skip specific components by names as that will at least provide an option for users who are struggling with incompatible components

nathan-gage:
will it be possible to resolve #611 with this PR -- i.e., not trying to name <Fragment/> from headless?

0Calories:
Hi @nathan-gage, unfortunately I don't think this would be able to help with that Headless UI issue. It's possible that you could include `React.Fragment` as a component name to be skipped, but I can't confirm whether that would work or not

0Calories:
@lforst This is good to go for a review now!

packages/bundler-plugin-core/src/types.ts
```diff
@@ -319,6 +319,15 @@ export interface Options {
...
     enabled?: boolean;
+    /**
+     * A list of strings representing the names of components to ignore. The plugin will not apply `data-sentry` annotations on the DOM element for these components.
+     */
+    ignoreComponents?: string[];
+    // TODO: Support this option in the future
+    /**
+     * A list of strings representing local paths to files to ignore. The plugin will not apply `data-sentry` annotations on components in these files
+     */
+    // ignoreFiles?: string[];
```
lforst, line 330: I would just remove this comment and create an issue instead.

packages/dev-utils/src/generate-documentation-table.ts
```diff
@@ -390,6 +390,21 @@ type IncludeEntry = {
...
+        supportedBundlers: ["webpack", "vite", "rollup"],
+      },
+      // TODO: Support this option in the future
+      // {
+      //   name: "ignoreFiles",
+      //   type: "string[]",
+      //   fullDescription:
+      //     "A list of strings representing local paths to files to ignore. The plugin will not perform any annotations on components in these files.",
+      //   supportedBundlers: ["webpack", "vite", "rollup"],
+      // },
```
lforst, line 407: Same as above

packages/babel-plugin-component-annotate/README.md
```diff
@@ -47,6 +47,14 @@ Using pnpm:
 pnpm add @sentry/babel-plugin-component-annotate --save-dev
 ```
 
+## Options
+
+### `ignoreComponents`
```
lforst, line 52: Super nit but I think my logic is sound: I think I would probably call it `ignore*d*Components`. `ignoreComponents` sounds like imperative and that it would accept a boolean rather than a list of components.

lforst (approved):
In lgtm, I think I would call it `ignoredComponents` though. Maybe a test or two would also be dope.

0Calories:
Thanks for the review @lforst! Agreed on the naming of `ignoredComponents`. There's also a test I added but it's buried in the diffs because of all the obsolete tests I deleted

Issue #551, opened 2024-06-18
Vite plugin logging to stdout instead of stderr


brynary:
### Environment

Vite, Remix

### Steps to Reproduce

1. Install `@sentry/vite-plugin` to send source maps and release notifications from CI/CD to Sentry
2. Intentionally do not store a Sentry auth token _locally_, only on CI/CD
3. Install [Knip](https://knip.dev/), a tool to find unused code
4. Run Knip with `npx knip --reporter json | jq .`

### Expected Result

Knip writes valid JSON to stdout. Any output from Knip would go to stderr not stdout because it is not the primary output of the execution.

### Actual Result

Sentry sends the following warning to stdout:

```[sentry-vite-plugin] Warning: No auth token provided. Will not upload source maps. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/```

As a result of this warning going to stdout, the output from `knip` is invalid JSON.

### Thoughts

- I am operating under the assumption that it is reasonable for us to only use a Sentry token in CI/CD, and not use one locally. Is that reasonable?
- Perhaps there is a way to explicitly tell the Vite plugin that it should not be trying to upload source maps, so it can understand there is no need for the warning
- Given that there are any number of log lines / warnings that the vite plugin could potentially emit, could the vite plugin logger be set to log to stderr instead of stdout?

lforst:
Hi, only logging to stderr seems reasonable to me.

> I am operating under the assumption that it is reasonable for us to only use a Sentry token in CI/CD, and not use one locally. Is that reasonable?

Correct. You can potentially even disable/remove the plugin in dev. It has very little gain in dev as of now, maybe even slightly slowing down your dev server.

> Perhaps there is a way to explicitly tell the Vite plugin that it should not be trying to upload source maps, so it can understand there is no need for the warning

You can set `disable: true` to disable the plugin.

brynary:
Thank you. Adding `disable: !process.env.SENTRY_AUTH_TOKEN` to our config solved the immediate problem

s1gr1d:
Since this was completed in https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/624, issues with other framworks occured and we had to change it again: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/652

However, this should not interfere with knip.

PR #624, opened 2024-10-28
chore: Log only to stderr

onurtemizkan:
Resolves: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/551

Switches logger to use `stderr`, not to interfere with `stdout`.


kaiyoma:
This seems like wrong semantics to me, and is causing a regression: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/650

PR #632, opened 2024-11-29
deps: Bump `@sentry/cli` to `2.39.1` and require specific version

lforst:
We recently had a regression in Sentry CLI that caused uploaded source maps to not be tagged with debug IDs which is mission critical.

Since we cannot allow buggy versions to float around we should pin the version.


andreiborza (approved):
😬

PR #636, opened 2024-12-05
feat(core): Write module injections to `globalThis`

lforst:
Some runtimes like QuickJS only have `globalThis`.


Lms24 (approved):
Sounds reasonable to me! Do we also need to adjust something in Sentry CLI?

lforst:
> Sounds reasonable to me! Do we also need to adjust something in Sentry CLI?

absolutely!

PR #644, opened 2025-01-06
fix: Use correct replacement matcher for `bundleSizeOptimizations.excludeTracing`

jdelStrother:
This flag was having no effect - the correct name is `__SENTRY_TRACING__`, not `__SENTRY_TRACE__`
(see, eg, https://github.com/getsentry/sentry-javascript/blob/8dd8ac556bb429f57ab77475575fbc6520191cdb/packages/core/src/utils/hasTracingEnabled.ts#L5)


lforst (approved):
oh dang thanks for catching that. Just checked and it seems like we never called it `__SENTRY_TRACE__` so this should be safe to do without any sort of backwards compat.

PR #646, opened 2025-01-07
fix!: Wrap injected code in block-statement to contain scope

### LINKED ISSUE #645: Injected global variable `_global` can clash with user code
We inject code that creates a new variable called `_global`:

https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/b2a0529ce11511ac4e54563af343d3e7c519f24b/packages/bundler-plugin-core/src/utils.ts#L317-L329

There is a chance this can clash with user code if that variable name is used.

We should either chose a variable name that is more unique and less likely to clash with user code or wrap this code block in a closure so that the scope is private.

timfish:
- Closes #645

### DIFF
--- a/packages/bundler-plugin-core/src/utils.ts
+++ b/packages/bundler-plugin-core/src/utils.ts
@@ -314,7 +314,7 @@ export function generateGlobalInjectorCode({
 }) {
   // The code below is mostly ternary operators because it saves bundle size.
   // The checks are to support as many environments as possible. (Node.js, Browser, webworkers, etc.)
-  let code = `
+  let code = `(function () {
     var _global =
       typeof window !== 'undefined' ?
         window :
@@ -334,6 +334,8 @@ export function generateGlobalInjectorCode({
     code += `
       _global.SENTRY_BUILD_INFO=${JSON.stringify(buildInfo)};`;
   }
+
+  code += '})();';
 
   return code;
 }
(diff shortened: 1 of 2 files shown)


lforst (approved):
Would it maybe be enough to just wrap the snippet in `{...}`. May save some bytes.

timfish:
> Would it maybe be enough to just wrap the snippet in `{...}`. May save some bytes.

That [doesn't work with `var`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_var_or_function_declaration_in_non-strict_mode) because of scoping rules. We would need to use `let` or `const` in this case.

That means that this block doesn't work/gets removed too:
https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/b2a0529ce11511ac4e54563af343d3e7c519f24b/packages/bundler-plugin-core/src/utils.ts#L346-L347

packages/bundler-plugin-core/src/utils.ts
```diff
@@ -344,7 +346,7 @@ export function generateModuleMetadataInjectorCode(metadata: any) {
   // The checks are to support as many environments as possible. (Node.js, Browser, webworkers, etc.)
   // We are merging the metadata objects in case modules are bundled twice with the plugin
   return `{
-  var _sentryModuleMetadataGlobal =
+  const _sentryModuleMetadataGlobal =
```
timfish, line 349: I fixed this too because [block scoping rules for `var`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_var_or_function_declaration_in_non-strict_mode) means that it wasn't working as expected

lforst: I am all for this but wondering whether we should cut a major in case somebody uses the bundler plugins for apps in an env where const isn't a thing yet.

PR #647, opened 2025-01-08
ref(core): Rename release management plugin name

Lms24:
- Renames the release management plugin name from `sentry-debug-id-upload-plugin` to `sentry-release-management-plugin`
- adds simple tests to catch renamings/regressions since they have a potential effect on SDKs like `@sentry/sveltekit`


lforst (approved):
Interesting PR description 😄 - otherwise lgtm

Issue #345, opened 2023-07-07
ng build stuck after successful build since sentryWebpack v2  


johanneszelger:
Since i upgraded from the plugin v1 to v2, my angular build gets stuck after a successfull build:

`Build at: 2023-07-07T10:40:52.046Z - Hash: cc2971c88cd12369 - Time: 32093ms`

just won't end the process after this line. When i remove the plugin, the build ends normally.  I tried the plugin versions 2.4 2.3 2.0, all unsuccessfull.

Lms24:
Hi, can you please share your plugin config? Do you get console output from the plugin regarding source maps upload or how many files were uploaded?

johanneszelger:
Sure, my config is:

`plugins = [
  new webpack.DefinePlugin({
    "process.env.RELEASE": JSON.stringify(release),
    "process.env.BUILD": JSON.stringify(build)
  })
]

if (process.env.SENTRY_AUTH_TOKEN !== undefined && process.env.SENTRY_AUTH_TOKEN !== "") {
  console.log("adding sentry webpack because auth token was found, release: ", release)
  plugins.push(sentryWebpackPlugin({
    url: "url",
    org: "org",
    project: "frontend",

    // Auth tokens can be obtained from https://sentry.io/settings/account/api/auth-tokens/
    // and need the `project:releases` and `org:read` scopes
    authToken: process.env.SENTRY_AUTH_TOKEN,

    // Specify the directory containing build artifacts
    // By default, Angular emits build files in the `dist` directory but your
    // configuration may vary
    // sourcemaps: {
    //   assets: ["./dist/Incentrium"]
    // },

    // Optionally uncomment the line below to override automatic release name detection
    release: {
      name: release,
      finalize: false,
      cleanArtifacts: true,
      dist: build
    }
  }))
} else
  console.log("skipping sentry webpack because auth token was not found")

module.exports = {
  devtool: "source-map", // Source map generation must be turned on
  plugins: plugins,
  watch: false,
};`

And yes i get output from the plugin, all sourcemaps are uploaded successfully, they are also visible on sentry. Everything is working - its just that my build pipeline gets stuck after the successfull build. If i cancel the process manually, everything works as expected.

MitkoTschimev:
We have the same issue. It just doesn't end the webpack process

MitkoTschimev:
I was able to pin it down to the dependency unplugin. 
For a simple plugin with this dep on the newest version shows the same behavior

[...]

PR #653, opened 2025-01-13
fix(webpack): Ensure process exits when done

Lms24:
This PR is heavily inspired by https://github.com/unjs/unplugin/issues/323#issuecomment-2208626770 where the suggestion is made to exit the process in the `done` hook of the webpack compiler. We have multiple reports of our webpack plugin stalling angular builds. This came up again recently so I thought we could give this fix a try.

Full disclosure: I cannot reproduce the stall posted in https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/345, but from my limited local testing, this at least doesn't seem to break anything. However, exiting the process is still scary, so not sure if we should really do it. I guess with TanStack, there's at least precedence for this.

Reviewers - WDYT?

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -33,7 +33,10 @@ interface SentryUnpluginFactoryOptions {
   componentNameAnnotatePlugin?: (ignoredComponents?: string[]) => UnpluginOptions;
   moduleMetadataInjectionPlugin: (injectionCode: string) => UnpluginOptions;
   debugIdInjectionPlugin: (logger: Logger) => UnpluginOptions;
-  debugIdUploadPlugin: (upload: (buildArtifacts: string[]) => Promise<void>) => UnpluginOptions;
+  debugIdUploadPlugin: (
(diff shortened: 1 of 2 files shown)


Lms24:
Probably related: https://stackoverflow.com/a/74368627

smeubank:
if this gets merged let's update the related [customer case](https://www.notion.so/sentry/thirdPartyErrorFilterIntegration-CLI-Support-Plugin-issue-in-Angular-87824a9d0d8346419947a1827ad5b2cc)

andreiborza (approved):
I'd say let's try, just concerned if people have plugins running after ours but since we specifically instruct people to add the bundler plugin as last plugin it should be fine.

lforst (approved):

Lms24:
okay sgtm. We can always revert!

lesderid:
This is breaking my build on Next.js 15.

lforst:
@lesderid please open an issue with reproduction steps. Thanks!

Lms24:
@lesderid version 3.1.0 should fix this (see #662)

Issue #651, opened 2025-01-09
Release a new major version


timfish:
We have some potential breaking changes to land that impact platform support 
- #610 (uses `globalThis`)
- #646 (uses `const` rather than `var`)

Remove deprecated APIs:
- #654

PR #654, opened 2025-01-14
chore: Remove deprecated options and update changelog for v3 major

timfish:
- Ref #651

Do you want to remove `injectBuildInformation` from `_experiments` too? It hasn't been deprecated yet 🤔

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,26 @@
 
 - "You know what they say ‘Fool me once, strike one, but fool me twice… strike three.’" — Michael Scott
 
+## 3.0.0
+
+### Breaking Changes
+
+- Code injected into bundles now uses:
+
+  - `const` which was added in ES6 (ES2015) (#646)
+  - `globalThis` which was added ES2020 but can be polyfilled (#610)
+
+- Deprecated configuration options have been removed:
+  - `deleteFilesAfterUpload` - Use `filesToDeleteAfterUpload` instead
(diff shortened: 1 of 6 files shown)


lforst:
> Do you want to remove injectBuildInformation from _experiments too? It hasn't been deprecated yet 🤔

Yes! We don't need a deprecation I think.

packages/dev-utils/src/generate-documentation-table.ts
```diff
@@ -412,18 +412,10 @@ type IncludeEntry = {
...
-      "Options that are considered experimental and subject to change. This option does not follow semantic versioning and may change in any release.",
-    children: [
-      {
-        name: "injectBuildInformation",
-        type: "boolean",
-        fullDescription:
-          "If set to true, the plugin will inject an additional `SENTRY_BUILD_INFO` variable. This contains information about the build, e.g. dependencies, node version and other useful data.\n\nDefaults to `false`.",
-      },
-    ],
+      "If set to true, the plugin will inject an additional `SENTRY_BUILD_INFO` variable. This contains information about the build, e.g. dependencies, node version and other useful data.\n\nDefaults to `false`.",
```
lforst, line 418: I was just able to visualize this a bit better now that you changed it and realized that we should probably still leave `injectBuildInformation` in experimental state. Sorry for all the back and fourth!

timfish: No problem, just a quick revert!

lforst (approved):

Issue #344, opened 2023-07-13
[Vite] No auth token provided in dev/watch mode warning


cleptric:
On https://docs.sentry.io/platforms/javascript/guides/vue/sourcemaps/uploading/vite/, we mention

> The Sentry Vite plugin doesn’t upload source maps in watch-mode/development-mode. We recommend running a production build to test your configuration.

However, running `vite`, I get the following output.

```bash
❯ npm run dev

> dev
> vite

[sentry-vite-plugin] Warning: No auth token provided. Will not create release. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
[sentry-vite-plugin] Warning: No auth token provided. Will not upload source maps. Please set the `authToken` option. You can find information on how to generate a Sentry auth token here: https://docs.sentry.io/api/auth/
```

owenjs:
It seems:

> Vite doesn't load .env files by default

So I needed to follow Vite's guide on how to use [Environment variables within their config file](https://main.vitejs.dev/config/#using-environment-variables-in-config)

Not sure if this has changed recently but this plugin's example is wrong.

lforst:
@owenjs I don't think any bundler loads .env files by default. I also don't think our example is wrong. If you wanna use vite's way of accessing environment configuration you have to do [that](https://main.vitejs.dev/config/#using-environment-variables-in-config), otherwise (as it says in those docs) "Environmental Variables can be obtained from `process.env` as usual".

@cleptric thanks for raising this, will fix.

jonrandahl:
For future visitors,

> Using `loadEnv` did the trick https://main.vitejs.dev/config/#using-environment-variables-in-config 

Following [franklinjavier](https://github.com/franklinjavier)'s approach is how I got this to work in my dev instance, if it helps anyone else then job done! 😁 

Logging out `env` allowed me to see all of my vars were loaded but I didn't need to use `process` or `import.meta` to access them pass them into my configuration:

```js
export default defineConfig(({ command, mode }) => {
  // Load env file based on `mode` in the current working directory.
  // Set the third parameter to '' to load all env regardless of the
  // `VITE_` prefix.
  const env = loadEnv(mode, process.cwd(), '')
  // console.log({env})

  return {
    plugins: [
      vue(),
      // Put the Sentry vite plugin after all other plugins
      sentryVitePlugin({
        org: env.SENTRY_ORG,
[...]

PR #666, opened 2025-01-27
fix(core): Check for dev mode via `process.env.NODE_ENV`

Lms24:
perfect PR id 😈 

This PR improves the check for if the plugin is called in dev/watch mode or for a production build. In case of the former, we return a no-op plugin. This was already the [expected behaviour](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/vite/#configuration) previously but due to the unplugin meta context only holding the `watchmode` flag in theory but not in practise, we'd always return the actual plugins.

This should also remove the log output in dev mode.

closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/344
probably also closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/520 (I'll leave a reply to get confirmation)

cc @cleptric

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -84,6 +84,11 @@ export function sentryUnpluginFactory({
       debug: userOptions.debug ?? false,
     });
 
+    // Not a bulletproof check but should be good enough to at least sometimes determine
+    // if the plugin is called in dev/watch mode or for a  prod build. The important part
+    // here is to avoid a false positive. False negatives are okay.
+    const isDevMode = process.env["NODE_ENV"] === "development";
+
     try {
       const dotenvFile = fs.readFileSync(
         path.join(process.cwd(), ".env.sentry-build-plugin"),
@@ -106,7 +111,7 @@ export function sentryUnpluginFactory({
 
     const options = normalizeUserOptions(userOptions);
 
-    if (unpluginMetaContext.watchMode || options.disable) {
+    if (isDevMode || options.disable) {
       return [
         {
           name: "sentry-noop-plugin",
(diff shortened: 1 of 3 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -274,7 +279,7 @@ export function sentryUnpluginFactory({
         "Release injection disabled via `release.inject` option. Will not inject release."
       );
     } else if (!options.release.name) {
-      logger.warn(
+      logger.debug(
```
Lms24, line 282: I also took the liberty to relax the log level here. Releases aren't required anymore for debugId-based upload and I've seen build logs this and the log below were added (in fact multiple times due to multi-build-setups of meta frameworks like SvelteKit and Nuxt)

Lms24:
Some more context on the `watchmode` flag: 

Unplugin [union-types](https://github.com/unjs/unplugin/blob/a5fe8fe418064b2ca9cdecb2c8ddd1a6c2fe36fe/src/types.ts#L88-L100) the `UnpluginMetaContext` with Rollup's `PluginMetaContext`. The `watchmode` flag comes from Rollup's type. However, when the plugin factory function is called, the meta object is only populated with a small subset of properties--mostly `framework`: https://github.com/unjs/unplugin/blob/a5fe8fe418064b2ca9cdecb2c8ddd1a6c2fe36fe/src/vite/index.ts#L9-L12

packages/bundler-plugin-core/src/index.ts
```diff
@@ -106,7 +111,7 @@ export function sentryUnpluginFactory({
 
     const options = normalizeUserOptions(userOptions);
 
-    if (unpluginMetaContext.watchMode || options.disable) {
```
lforst, line 109: I know this is technically not the goal of this PR but thinking about this logic more holistically we probably don't want to disable the entire plugin for dev mode because we have other things that may be valuable in dev mode, like react component annotations or tree shaking utilities (I guess for debugging purposes).

Should we limit this check to just anything that has a side-effect/is invoking Sentry CLI?

lforst (approved):

Issue #686, opened 2025-02-25
reactComponentAnnotation passes props to `Fragment`


richardasymmetric:
### Is there an existing issue for this?

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Which SDK are you using?

@sentry/nextjs

### SDK Version

9.2.0

### Framework Version

react@npm:19.0.0-rc-66855b96-20241106 next@npm:15.1.7 [1ff14]

### Link to Sentry event

https://flowtax.sentry.io/issues/6331780118/events/99036a9409274672af8e72c6295349ba/

### Reproduction Example/SDK Setup

I'm going to have a hard time creating a minimal reproduction since I'm still not entirely sure what's causing it, but I have narrowed it down. However, he's my sentry configs:

```javascript
// next.config.js
/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: true,
  poweredByHeader: false,

  async headers() {
    return [
      {
        source: "/:path*",
        headers: [
          {
            key: "x-content-type-options",
            value: "nosniff",
          },
          {
            key: "X-Frame-Options",
            value: "SAMEORIGIN",
          },
        ],
      },
    ];
  },

  serverExternalPackages: [
    "@prisma",
    "knex",
    "@aws-sdk/client-s3",
    "@aws-sdk/s3-request-presigner",
  ],

  compiler: {
    removeConsole:
      process.env.NODE_ENV === "production"
        ? {
            exclude: ["error", "warn"],
          }
        : false,
  },
};

// Injected content via Sentry wizard below

import { withSentryConfig } from "@sentry/nextjs";

function wrapper(nextConfig, sentryConfig) {
  if (process.env.TURBOPACK && !process.env.VERCEL_ENV) {
    console.info("Skipping Sentry setup while TurboPack is in use in development");
    return nextConfig;
  }
  return withSentryConfig(nextConfig, sentryConfig);
}

export default wrapper(nextConfig, {

  org: org here,
  project: project here,
  sentryUrl: "https://sentry.io/",

  silent: !process.env.CI,

  widenClientFileUpload: true,

  reactComponentAnnotation: {
    enabled: true,
  },

  tunnelRoute: "/monitoring",

[...]

PR #687, opened 2025-02-26
feat: Ignore `Fragment` components by default for react component annotation

lforst:
Fixes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/686

I think there are like 0 cases where we would actually want to annotate fragments as they do not manifest in dom elements in any case.

### DIFF
--- a/packages/babel-plugin-component-annotate/src/index.ts
+++ b/packages/babel-plugin-component-annotate/src/index.ts
@@ -57,6 +57,8 @@ interface AnnotationPluginPass extends PluginPass {
 
 type AnnotationPlugin = PluginObj<AnnotationPluginPass>;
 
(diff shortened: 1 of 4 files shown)


0Calories:
Thanks @lforst 🙏 
Ideally we should have the default skipped components set in the babel plugin rather than the Sentry bundler plugins. Otherwise standalone users of the component annotation plugin will not have the changes in this PR

btpnlsl:
Would [this check](https://github.com/getsentry/sentry-javascript-bundler-plugins/blame/ee73414589a3341c4a4a8ec8efa3116d838e33f8/packages/babel-plugin-component-annotate/src/index.ts#L453) in `isReactFragment` (which is called by the `applyAttributes` method) not already filter out adding attributes to Fragments already?

PR #690, opened 2025-02-28
feat(annotation): Handle JSX member expressions

chargome:
This pr adds functionality for ignoring components like `Tab.Group` which are commonly used e.g. in HeadlessUI.

closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/686

### DIFF
--- a/packages/babel-plugin-component-annotate/src/index.ts
+++ b/packages/babel-plugin-component-annotate/src/index.ts
@@ -533,6 +533,29 @@ function getPathName(t: typeof Babel.types, path: Babel.NodePath): string {
     return name.name.name;
   }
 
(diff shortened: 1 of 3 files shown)


lforst (approved):

packages/babel-plugin-component-annotate/src/index.ts
btpnlsl, line 1: @chargome I believe with your changes to `getPathName` that the if-block in lines 458 to 494 is duplicate code.  `getPathName` should now return the compound names for `JSXMemberExpression` nodes instead of 'unknown'.

chargome: Yeah I'd still leave it as a fallback

PR #692, opened 2025-03-05
feat: Default to automatically setting commits on release

lforst:
It's just way better UX if we default to automatically setting commits on releases if Sentry CLI is able to detect them.

We just have to jump through some hoops to not throw by default but throw if people manually configured stuff.

### DIFF
--- a/packages/bundler-plugin-core/src/plugins/release-management.ts
+++ b/packages/bundler-plugin-core/src/plugins/release-management.ts
@@ -13,7 +13,7 @@ interface ReleaseManagementPluginOptions {
   shouldCreateRelease: boolean;
   shouldFinalizeRelease: boolean;
   include?: string | IncludeEntry | Array<string | IncludeEntry>;
-  setCommitsOption?: SentryCliCommitsOptions;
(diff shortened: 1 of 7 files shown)


packages/bundler-plugin-core/src/plugins/release-management.ts
```diff
@@ -86,8 +87,20 @@ export function releaseManagementPlugin({
...
+        if (setCommitsOption !== false) {
+          try {
+            await cliInstance.releases.setCommits(releaseName, setCommitsOption);
+          } catch (e) {
+            // isDefault being present means that the plugin defaulted to `{ auto: true }` for the setCommitsOptions, meaning that wee should not throw when CLI throws because there is no repo
+            if (!("isDefault" in setCommitsOption)) {
+              throw e;
+            } else {
+              logger.debug(
+                "An error occurred setting commits on release (this message can be ignored unless you commits on release are desired):",
```
chargome, line 99:
```suggestion
                "An error occurred setting commits on release (this message can be ignored unless your commits on release are desired):",
```

chargome (approved):

PR #698, opened 2025-03-11
fix: only show missing authToken warning if `release.create` is true

### LINKED ISSUE #536: Dont log warnings when certain features are disabled
Ref https://github.com/getsentry/sentry-javascript/issues/12152

eni4sure:
Fixes #536

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -347,7 +347,7 @@ export function sentryUnpluginFactory({
       );
     } else if (isDevMode) {
       logger.debug("Running in development mode. Will not create release.");
-    } else if (!options.authToken) {
+    } else if (!options.authToken && options.release.create) {
       logger.warn(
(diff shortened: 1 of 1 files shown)


lforst:
Hi, unfortunately we cannot merge this PR as-is because it will actually change whether the release management plugin is added or not, which controls a bunch of other things too. We would need to be a little bit smarter about this.

eni4sure:
@lforst That makes sense.  I pushed an update to explicitly check if both `release.create` and `release.finalize` is set to false to disable releases. That should avoid any unintended changes.

Lms24:
Hey @eni4sure 👋 Really sorry that this PR went unnoticed for such a long time. We did quite a few refactors in the meantime. Are you still up to get this PR over the finish line? If so, could you rebase it to the latest `main` commit and re-check if the fix still applies correctly? If not, that's of course fine, too. Again, sorry for the delayed response!

eni4sure:
hey @Lms24, haven't tested but would close this for now.

PR #700, opened 2025-03-18
feat(core): Use path instead of debug IDs as artifact names for debug ID upload

lforst:
This PR changes our "upload preparation" logic in the debug ID upload to not emit files named after their debug IDs, but rather after their actual file-names.

Files that would have previously been uploaded as `~/0-96b0eb2b-013a-5640-8ffc-4fd8465bb570.js` would now be uploaded as `~/0/out/rollup/entrypoint1.mjs` which is slightly easier to debug.

We still need to do some kind of collision-prevention in case two files are named identically relative to cwd after normalization, so that's why there's still a `/{number}` prefix in the artifact name (it is the "bundle index" which is an incrementing counter).

### DIFF
--- a/packages/bundler-plugin-core/src/debug-id-upload.ts
+++ b/packages/bundler-plugin-core/src/debug-id-upload.ts
@@ -232,24 +232,7 @@ export async function prepareBundleForDebugIdUpload(
     return;
   }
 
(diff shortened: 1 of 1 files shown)


lforst:
@codecov-ai-reviewer review

Lms24 (approved):
This is a great change!

PR #711, opened 2025-03-31
feat: Only do automatic commit association for Vercel production environments

lforst:
Just noticed this is extremely spammy when you have your git email connected with Sentry.

### DIFF
--- a/packages/bundler-plugin-core/src/options-mapping.ts
+++ b/packages/bundler-plugin-core/src/options-mapping.ts
@@ -48,7 +48,11 @@ export function normalizeUserOptions(userOptions: UserOptions) {
       process.env["VERCEL"] &&
       process.env["VERCEL_GIT_COMMIT_SHA"] &&
       process.env["VERCEL_GIT_REPO_SLUG"] &&
-      process.env["VERCEL_GIT_REPO_OWNER"]
+      process.env["VERCEL_GIT_REPO_OWNER"] &&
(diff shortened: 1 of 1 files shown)


packages/bundler-plugin-core/src/options-mapping.ts
```diff
@@ -48,7 +48,11 @@ export function normalizeUserOptions(userOptions: UserOptions) {
       process.env["VERCEL"] &&
       process.env["VERCEL_GIT_COMMIT_SHA"] &&
       process.env["VERCEL_GIT_REPO_SLUG"] &&
-      process.env["VERCEL_GIT_REPO_OWNER"]
+      process.env["VERCEL_GIT_REPO_OWNER"] &&
+      // We only want to set commits for the production env because Sentry becomes extremely noisy (eg on slack) for
+      // preview environments because the previous commit is always the "stem" commit of the preview/PR causing Sentry
+      // to notify you for other people creating PRs.
+      process.env["VERCEL_TARGET_ENV"] === "production"
```
Lms24, line 55: Is there a way that users can overwrite this env variable for their production build?

Lms24 (approved):
Sounds reasonable to me!

packages/bundler-plugin-core/src/options-mapping.ts
lforst: Had the same thought: I think we just don't care. I don't think many people would change their prod env name at all.

PR #713, opened 2025-04-01
feat(core): Expose bundler plugin primitives via `createSentryBuildPluginManager`

lforst:
This PR adds a `createSentryBuildPluginManager` function to the `@sentry/bundler-plugin-core` package that exposes most of the primitives that are necessary to build a Sentry bundler plugin, or a Sentry plugin for JS build tooling.

- Why do we need a "manager" instance for this and not just expose the primitives directly?
	- A lot of the logic in the bundler plugins comes from the options that are passed in. If we were to expose the primitives directly, the option normalization would always have to be handled by the consuming package and that would get very cumbersome and prone to errors and inconsistencies.

This PR leaves the core package in an extremely desolate state. Thisis because it required major refactorings but I didn't want to clean stuff up in the same PR because it would get awful to review.

### DIFF
--- a/packages/integration-tests/fixtures/telemetry/telemetry.test.ts
+++ b/packages/integration-tests/fixtures/telemetry/telemetry.test.ts
@@ -39,122 +39,112 @@ test("rollup bundle telemetry", async () => {
   // Ensure the session gets closed
   process.emit("beforeExit", 0);
 
-  expect(gbl.__SENTRY_INTERCEPT_TRANSPORT__).toEqual([
-    // Fist we should have a session start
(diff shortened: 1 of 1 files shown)


AbhiPrasad:
What's the motivation? To eventually move off of unplugin?

lforst:
> What's the motivation? To eventually move off of unplugin?

It's basically needed to support Turbopack in Next.js. See https://github.com/getsentry/sentry-javascript/pull/15779.

Turbopack doesn't have a plugin system so we need to access the primitives from the Next.js layer for now.

AbhiPrasad (approved):

PR #715, opened 2025-04-02
feat(webpack): Add `sentry/webpack-plugin/webpack5` export for webpack 5.1+ and compatible environments

krystofwoldrich:


packages/webpack-plugin/rollup.config.webpack5.js
lforst, line 1: Any reason we need two rollup configs? Can we do it in one?

krystofwoldrich: We can do it in one if it's okay that the `index.js` and `webpack5.js` are not bundles.

I merged the config here: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/715/commits/ac046ccae7ffc2fa9bfde25579b113a26cba97a5

lforst: jup that seems way better. Thanks!

packages/webpack-plugin/src/plugin.ts
```diff
@@ -0,0 +1,227 @@
...
+          raw: true,
+          include: /\.(js|ts|jsx|tsx|mjs|cjs)(\?[^?]*)?(#[^#]*)?$/,
+          banner: injectionCode,
+        })
+      );
+    },
+  });
+}
+
+export function sentryWebpackUnpluginFactory({
```
lforst, line 198: Let's add a comment here why this takes the banner/define plugins. ie: To not have to actually depend on webpack.

lforst (approved):
Nice solution!

PR #722, opened 2025-04-10
feat: Use existing Debug IDs from bundles

### LINKED ISSUE #710: Cater for when bundlers have already injected Debug IDs
Recent versions of Webpack, Rollup and Vite can now inject debug IDs into source and sourcemaps files.

If source and sourcemap already contains a debug ID, the bundler plugins should use these existing IDs and should only inject the polyfill.

timfish:
- Closes #710

Will need the repository Node version bumping to v18 before tests can be added.


timfish:
See here:
- https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/710#issuecomment-2832030837

Issue #727, opened 2025-04-16
Supply correct return type for `sentryVitePlugin`


Lutymane:
The factory function actually returns an array of plugins which caused issues with plugin management with an expectation it returns a single plugin. It should be noted either in docs, or properly typed to avoid issues for other users.

https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/9c564544b72caaea48205ce77ba6b4ddf0af09d2/packages/vite-plugin/src/index.ts#L77

chargome:
Hey @Lutymane, thanks for writing in – we'll have a look!

PR #728, opened 2025-04-17
feat(vite): Update return type of vite plugin

chargome:
closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/727


chargome:
Will leave this as a draft for now as this will introduce a breaking change
cc @Lms24

PR #732, opened 2025-04-30
feat(core): Add hook to customize source map file resolution

### LINKED ISSUE #731: Add hook to customize source map file resolution
The sentry plugin [currently uses heuristics to locate sourcemap files for a given bundle file](https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/2d796d7/packages/bundler-plugin-core/src/debug-id-upload.ts#L103), but in some build configurations this is insufficient. Specifically if the sourcemaps are being written to a separate directory. It would be good to have a hook, similar to `sourcemaps.rewriteSources`, that would allow you to manually resolve the `.map` files.

In our specific webpack build we're writing our sourcemaps into a separate directory so they can be bundled up and hosted separately. To do this we configure the `SourceMapDevToolPlugin` with a custom filename pattern and `//# sourceMappingURL` line. However, the sentry plugin doesn't know about or understand any of these configs, so it's unable to locate the source map files.

Here's an example of how our solution might look:

```js
module.exports = {
  devtool: false,
  plugins: [
    new webpack.SourceMapDevToolPlugin({
      filename: `../sourcemaps/[file].map`,
      append: `\n//# sourceMappingURL=${SOURCE_MAP_HOST}/[file].map`,
    }),
    sentryWebpackPlugin({
      sourcemaps: {
        resolveSourceMap(bundlePath, sourceMappingUrl) { 
          return sourceMappingUrl.replace(SOURCE_MAP_HOST, SOURCEMAP_DIR)
        }
      },
    })
  ]
}
```

thecodewarrior:
I wasn't able to implement full integration tests for the sourcemap changes, but there's a smoke test in the playground and I added full unit tests for the function that performs the file resolution.

One of the only changes I made to the existing behavior is using `url.fileUrlToPath()` when the `sourceMappingURL` is a `file://` url. I added some additional debug logging to that method as well. Now when it fails to locate a sourcemap it prints all the locations it tried, and when it succeeds it prints out the source map it chose.

Fixes #731


BoomchainLabs (approved):

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -114,61 +113,62 @@ function addDebugIdToBundleSource(bundleSource: string, debugId: string): string
...
 
-    let isUrl;
-    let isSupportedUrl;
+  const searchLocations: string[] = [];
+
+  if (resolveSourceMapHook) {
+    const customPath = await resolveSourceMapHook(bundlePath, sourceMappingUrl);
+    if (customPath) {
+      searchLocations.push(customPath);
+    }
```
andreiborza, line 131: We should add a debug log when the custom path doesn't exist. While a fallback to the next location does log out which path was ultimately used, it might be hard to understand what's going on when supplying the hook but making a mistake, wdyt?

andreiborza (commented):
Hi, thanks so much for contributing this and the nice test suite!

packages/bundler-plugin-core/src/types.ts
```diff
@@ -124,8 +124,17 @@ export interface Options {
...
+
+    /**
+     * Hook to customize source map file resolution.
+     *
+     * The hook is called with the absolute path of the build artifact and the value of the `//# sourceMappingURL=`
+     * comment, if present. The hook should then return an absolute path indicating where to find the artifact's
+     * corresponding `.map` file. If no path is returned or the returned path doesn't exist, the standard source map
+     * resolution process will be used.
+     */
+    resolveSourceMap?: ResolveSourceMapHook;
```
lforst, line 137: Let's not forget to add this option to https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/24191277f07f86488e92e24353305ee09d6a8c71/packages/dev-utils/src/generate-documentation-table.ts#L11 with a good description. Ideally, we also add a short explainer paragraph to the JS doc and the docs in what situations this is beneficial.

packages/bundler-plugin-core/src/debug-id-upload.ts
```diff
@@ -114,61 +113,62 @@ function addDebugIdToBundleSource(bundleSource: string, debugId: string): string
...
+      // noop
+    }
   }
 
   // This is just a debug message because it can be quite spammy for some frameworks
   logger.debug(
-    `Could not determine source map path for bundle: ${bundlePath} - Did you turn on source map generation in your bundler?`
+    `Could not determine source map path for bundle: \`${bundlePath}\`` +
+    ` - Did you turn on source map generation in your bundler?` +
+    ` (Attempted paths: ${searchLocations.map(e => `\`${e}\``).join(", ")})`
```
lforst, line 171: low key love this change.

packages/bundler-plugin-core/src/debug-id-upload.ts
thecodewarrior: That's a good idea. Having clear feedback of what your custom hook is generating would be helpful when writing the callback. Especially if we include the passed `bundlePath` and `sourceMappingUrl` in the message.

packages/bundler-plugin-core/src/types.ts
thecodewarrior: Thanks for pointing that out, I wasn't aware of that file. It might be good to add a note about it to the `CONTRIBUTING.md` file.

packages/bundler-plugin-core/src/debug-id-upload.ts
thecodewarrior: While I was addressing the other comments I made it so this logging statement also includes the `sourceMappingURL` value, so it should be clearer why it might be searching in specific locations

PR #744, opened 2025-05-21
chore: Add changelog generation script

Lms24:
Adds the same `yarn changelog` script as in the `sentry-javascript` repo to extract a pre-formatted commit list we can easily insert into our CHANGELOG entries. Trialed in #743

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "build:watch": "nx run-many --target=build:watch --all",
     "build:graph": "nx graph",
     "build:npm": "nx run-many --target=build:npm --all",
+    "changelog": "ts-node ./scripts/get-commit-list.ts",
     "check:types": "nx run-many --target=check:types --all",
     "clean": "nx run-many --target=clean --all",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "build:watch": "nx run-many --target=build:watch --all",
     "build:graph": "nx graph",
     "build:npm": "nx run-many --target=build:npm --all",
+    "changelog": "ts-node ./scripts/get-commit-list.ts",
```
andreiborza, line 16: Where's the script? :)

Lms24: classic forgot to `git add` mistake 😅

andreiborza (approved):
nice nice!

yarn.lock
```diff
@@ -11500,6 +11500,13 @@ rollup@2.77.0:
   optionalDependencies:
     fsevents "~2.3.2"
 
+rollup@2.79.2:
```
Lms24, line 11503: this just slipped in b/c I didn't rebase to latest main (dep upgrade via #736)

Issue #750, opened 2025-07-01
Regression in Vite plugin drops Windows ARM64 support


wojtekmaj:
Upgrading @sentry/vite-plugin from 2.10.3 to 2.23.0 caused a regression, dropping Windows ARM64 support.

* 2.10.3 relied on @sentry/bundler-plugin-core v2.10.3, which had @sentry/cli dependency specified as ^2.22.3
* 2.23.0 relies on @sentry/bundler-plugin-core v2.23.0, which has @sentry/cli dependency specified as 2.39.1 (WITHOUT the caret)

This caused @sentry/cli version to be DOWNGRADED when I updated Vite plugin - from 2.42.5 in my case to 2.39.1 (of course). This, in turn, removed Windows ARM64 support, causing my project to crash on install on Windows ARM64 machine.

PR #751, opened 2025-07-01
deps: Bump `@sentry/cli` to `^2.46.0`

mydea:
Also make it use `^` so we also get future updates as well.

Closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/750


mydea:
Needed to adjust an integration test that timed out because of auto-http-retry that was added in 2.46.0 (https://github.com/getsentry/sentry-cli/pull/2523/).

andreiborza (approved):

PR #753, opened 2025-07-07
feat(core): Don't add `debugIdUploadPlugin` when sourcemaps option is disabled

chargome:
closes https://github.com/getsentry/sentry-javascript/issues/16810


packages/integration-tests/package.json
```diff
@@ -47,4 +47,4 @@
   "volta": {
     "extends": "../../package.json"
   }
-}
+}
```
AbhiPrasad, line 50:
```suggestion
}

```

AbhiPrasad (approved):

packages/rollup-plugin/src/index.ts
```diff
@@ -45,6 +45,12 @@ function rollupDebugIdUploadPlugin(
   logger: Logger,
   createDependencyOnBuildArtifacts: () => () => void
 ): UnpluginOptions {
+  // Only for testing purposes
+  if (process.env["SENTRY_NODE_ENV"] === "test") {
```
Lms24, line 49: I think for rollup and vite, we could simply assert on the names of the returned plugins in some unit tests. For webpack, we probably need something like this indeed 🤔 so no strong objections to not doing it

Lms24 (approved):

Issue #755, opened 2025-07-09
Allow globs in uploadLegacySourcemaps


mattico:
I'm using vite and the sentryVitePlugin. My app uses web workers which are imported with `new Worker(new URL('./workerLineStats.ts', import.meta.url), { type: 'module' });`. Vite understands that and outputs `workerLineStats-LjpNa9wZ.js` and a `.map` file.

I'm using DebugIDs for the main app, but those don't work for the web worker files. Correct me if I'm wrong, I'm going off of a comment I left when I set this up long ago. I want to use `uploadLegacySourceMaps` for the web worker files, but it's hard because the filenames include a cache-busting hash.

If `uploadLegacySourcemaps` supported globs I could solve my problem easily with something like this:

```
uploadLegacySourcemaps: [
  'build/assets/workerLineStats-*.js.map'
]
```

chargome:
@mattico would generally just providing the asset path work for you?

```
uploadLegacySourcemaps: [
  'build/assets/'
]
```

Lms24:
@mattico thanks a lot for investigating and reporting back! I agree that we should try to support worker bundles as well. Are you by any chance able to share a reproducible example? No worries if not! I'm currently trying to reproduce this myself, though my implementation might of course differ from yours :D

> Am I missing something?

I will double-check the logic here -- on first glance, your options look fine. You might need to create a `release` as well, since legacy source maps are matched by `release` name. (Not 100% sure yet because this is ... well.. a legacy path but I'll confirm in a bit).

timfish:
Is there some way we detect when vite creates a chunk for a worker? If there is, maybe we could include that debugId with the parent that loads the worker? That way the debugID would already be where we need it?

timfish:
If we inject the same Sentry plugin instance into both the root and worker config, and have a way to track when it is called for a worker, can we track the debugIds for workers and include them with all other chunks? 

Just realised it's not that simple because our debugID injection uses `new Error().stack` to get the file URL and that wouldn't work in this case 🤔

```ts
const sentryVite =   sentryVitePlugin({
    org: "sentry-sdks",
    project: "javascript",
});

export default defineConfig({
  plugins: [sentryVite],
  worker: {
    plugins: () => [sentryVite.worker], // <- we'll know when the plugin is used for a worker...
  },
});
```

PR #756, opened 2025-07-10
feat(core): Allow globs for uploading legacy sourcemaps

chargome:
closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/755


Lms24:
@chargome given the progress of #755, let's close this PR for now? (if you think this would be a good change I'm also happy to merge it in a major)

Issue #670, opened 2025-02-03
Bundler plugin says upload successful, when Sentry CLI fails


szokeasaurusrex:
This issue was originally reported by @kubijo in the [Sentry CLI repository](https://github.com/getsentry/sentry-cli/issues/2373).

This issue report is based on the information provided in the [original issue](https://github.com/getsentry/sentry-cli/issues/2373); I did not try to reproduce the bug myself.

@kubijo, it would be helpful if you could fill in any missing information here, such as the version of the bundler plugin you are using, and any reproduction steps which are missing. 

### Environment

`"@sentry/webpack-plugin": "3.1.1"`, with Sentry CLI 2.39.1. Running on Ubuntu 24.10.

### Steps to Reproduce

1. Use the bundler plugin to upload sourcemaps to Sentry
2. When uploading, configure a non-existing project as the destination, so that Sentry CLI fails during the upload.
    - In the original issue, the Sentry CLI failed due to a 502 error. However, triggering any kind of failure in the Sentry CLI should likely cause the same behavior in the bundler plugin.

### Expected Result

Bundler plugin logs that an error prevented upload.

### Actual Result

Bundler plugin logs that sourcemap upload was successful:

```
[...]
> Bundled 42 files for upload
> Bundle ID: 95362a95-0674-5519-ad70-4fc1c6bb1d63
error: Project not found. Ensure that you configured the correct project and organization.

[...]

[sentry-webpack-plugin] Info: Successfully uploaded source maps to Sentry

```

kubijo:
```
"@sentry/webpack-plugin": "3.1.1",
```

kubijo:
For some more annoyances...
- even though I supplied `{ debug: true }` flag in plugin options, it prints `Add --log-level=[info|debug] or export SENTRY_LOG_LEVEL=[info|debug] to see more output.` on error.
- you depend on an old CLI here, can you keep it up to date?

szokeasaurusrex:
@kubijo we like to keep our issues limited in scope. So, please open a separate issue for the first item you have mentioned. 

@lforst has already bumped the Sentry CLI dependency to the latest version in #671. I believe he is planning to release a new version of the bundler plugin today.

jakeoeding:
I am observing the same issue in `@sentry/esbuild-plugin` v3.2.2

Example output snippet:
```
error: API request failed
 
Caused by:
    sentry reported an error: Invalid token (http status: 401)
 
[...]

PR #763, opened 2025-07-18
fix(core): Observe and handle Sentry CLI sourcemap upload failures

Lms24:
This PR bumps `@sentry/cli` to 2.49.0 which adds a new `rejectOnError` live mode setting to the `execute` and `releases.uploadSourceMaps` APIs. 

This new mode now makes the CLI reject the promise it returns if the CLI binary process exits with a non-zero exit code. In the previous live mode, it never rejected but only piped the stdio output. 

This should finally fix the "Successfully uploaded source maps" message when in fact we failed during upload. 

closes  https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/670


andreiborza (approved):
nice, good stuff!

PR #770, opened 2025-07-21
fix(webpack): make plugin inject ES5-friendly code

### LINKED ISSUE #769: `@sentry/webpack-plugin` Adds Code That Breaks Old Browsers
### Environment

Chromium 38, `@sentry/webpack-plugin` 3.5.0.

### Steps to Reproduce

1. Build the app with webpack and `@sentry/webpack-plugin` included in the list of plugins.
2. Run the app in an old browser.

### Expected Result

The app starts up without errors.

### Actual Result

A syntax error near `let` keyword is thrown upon app start.

### Additional Details

Sentry injects some code that has `let` in it. [`let` syntax](https://caniuse.com/let) is not supported by the old browsers.

grushetsky:
This PR fixes #769.

Sentry webpack plugin injects code that includes `let` keyword that was introduced with ES2015. [`let` syntax](https://caniuse.com/let) is not supported by some old browsers which are still relevant in certain cases.

The code is injected by the plugin at the final stages of bundling when webpack loaders that downgrade the syntax to the target environment (ES5) have already finished the processing. Therefore, `let` is kept in the bundle.

Changes:
* Substitute `let` for `var`.
* Wrap `var` variable declarations with an IIFE to keep the same scope as it was with `let`.

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -456,7 +456,7 @@ export function createComponentNameAnnotateHooks(ignoredComponents?: string[]): 
 }
 
 export function getDebugIdSnippet(debugId: string): string {
-  return `;{try{let e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof globalThis?globalThis:"undefined"!=typeof self?self:{},n=(new e.Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="${debugId}",e._sentryDebugIdIdentifier="sentry-dbid-${debugId}")}catch(e){}};`;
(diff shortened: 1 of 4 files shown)


Lms24 (commented):
Hey, thanks a lot for opening this PR! As outlined in #769, for now we'll close this. If we reconsider, I'll be happy to reopen and merge it.

UDPATE: talking within the team right now if we don't simply go forward with this solution

Lms24 (approved):
This looks reasonable to me! I ran the test snapshots on my end, in addition to adding a unit test for the release and build information snippet. Everything fine so far. 

However, after merging this, I'll follow up with a PR to "minify" the release and module metadata snippets. This shouldn't block you any further but it will be somewhat necessary for #761, since the added code will no longer go through minification because of the `renderChunk` hook timing.

grushetsky:
@Lms24, oh, adding an extra test is a good idea! Thank you for merging the fix!

PR #783, opened 2025-08-06
fix(debugId): Add guards for injected code to avoid errors

s1gr1d:
The injected code snippets where changed in those PRs:
- https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/770
- https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/761

This PR adds a try/catch guard around the snippet to avoid errors.

closes https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/770

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -431,7 +431,7 @@ export function createComponentNameAnnotateHooks(ignoredComponents?: string[]): 
 }
 
 export function getDebugIdSnippet(debugId: string): string {
-  return `;{try{(function(){var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof globalThis?globalThis:"undefined"!=typeof self?self:{},n=(new e.Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="${debugId}",e._sentryDebugIdIdentifier="sentry-dbid-${debugId}");})();}catch(e){}};`;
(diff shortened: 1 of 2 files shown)


packages/bundler-plugin-core/src/index.ts
```diff
@@ -431,7 +431,7 @@ export function createComponentNameAnnotateHooks(ignoredComponents?: string[]):
 }
 
 export function getDebugIdSnippet(debugId: string): string {
-  return `;{try{(function(){var e="undefined"!=typeof window?window:"undefined"!=typeof global?global:"undefined"!=typeof globalThis?globalThis:"undefined"!=typeof self?self:{},n=(new e.Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="${debugId}",e._sentryDebugIdIdentifier="sentry-dbid-${debugId}");})();}catch(e){}};`;
+  return `;{try{(function(){var e=("undefined"!=typeof window&&window)||("undefined"!=typeof global&&global)||("undefined"!=typeof globalThis&&globalThis)||("undefined"!=typeof self&&self)||{},n=(new e.Error).stack;n&&(e._sentryDebugIds=e._sentryDebugIds||{},e._sentryDebugIds[n]="${debugId}",e._sentryDebugIdIdentifier="sentry-dbid-${debugId}");})();}catch(e){}};`;
```
Lms24, line 434: what was the concern with the ternary chain here? Isn't it smaller in bundle size?

Lms24 (commented):
Which issue does this close? The "closes" currently just links to a previous PR :)

Also, I thought the snippet was safe enough to avoid the size hit of the try/catch but chances are high I missed something. So I think we should proceed with this PR.

timfish (approved):

s1gr1d:
Yeah, I'm also not 100% sure about this specific case but adding a try/catch here is a good idea in general.

PR #785, opened 2025-08-07
feat(core): Add flag for disabling sourcemaps upload

chargome:
Adds functionality so we can set `sourcemaps.disable : 'disable-upload'` which in turn still injects debugIds but does not upload them.

ref https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/784
ref https://linear.app/getsentry/issue/JS-267/slow-build-times-and-webpack-serialization-performance-warnings-when


packages/bundler-plugin-core/src/types.ts
```diff
@@ -93,9 +93,12 @@ export interface Options {
     /**
      * Disables all functionality related to sourcemaps.
      *
```
Lms24, line 95:
```suggestion
     * Disables all functionality related to sourcemaps if set to `true`.
     *
```

Lms24 (commented):
I'm wondering if we'll ever need a `"disable-inject"` option. At that point we're in the funny situation that three truthy values disable stuff while one false(y) value enables everything. But given we have the `disable` semantics already, this is better than flipping everything around again. 

So I think this is fine overall. I briefly thought if we should use an object instead (`{upload: boolean, inject: boolean}`) but this is equal in semantics and just differs in syntax. So not really worth changing. 

So let's leave it as-is but this needs a test before we merge it :)

Lms24 (approved):
Thanks for adding tests!

andreiborza (approved):

braddialpad:
Is this option documented anywhere? I only was able to find it by looking at the actual code.

PR #792, opened 2025-08-18
perf: use premove for build clean

43081j:
rimraf has grown rather large these days. We can now use `premove`, a much smaller utility.

Alternatively, we can just `fs.rm` in a script if we don't care for a CLI anymore.


Lms24 (approved):
Thanks!

PR #797, opened 2025-08-28
fix(webpack-plugin): forward headers option to sentry-cli

### LINKED ISSUE #796: 🐞 Bug: headers option does not set custom HTTP headers
### Description
According to the [plugin documentation](https://www.npmjs.com/package/@sentry/webpack-plugin#headers), the headers option should allow adding additional HTTP headers to outgoing requests. However, setting this option has no effect — no custom headers are added, and the debug logs do not appear with `--header` arguments being passed to `sentry-cli`.

### Environment

@sentry/webpack-plugin vesion: 4.1.1  
Framework: Vue 2 (via @vue/cli-service)
Node version: v24.5.0

### Steps to Reproduce

1. Add a `headers` option in the plugin config.
2. Run a build with `debug: true`. Preferably set the `SENTRY_LOG_LEVEL` to `debug`
3. Observe that the header is not present in debug logs and not sent in requests.

### Expected Result

- Custom headers (e.g., `x-identifier`) should be forwarded to `sentry-cli` and included in upload requests.
- The debug logs should reflect the `--header` flags.

### Observed Behavior
- The debug logs show the command executed by the plugin and some --header is  included.
- But no custom headers (those which are set in the `headers` option) are sent with the request.

### Actual Result

```
  INFO    2025-08-28 15:49:52.439394471 +03:30 sentry-cli was invoked with the following command line: ".../sentry-javascript-bundler-plugins/node_modules/@sentry/cli-linux-x64/bin/sentry-cli" "--header" "sentry-trace:xxxxxxx" "--header" "baggage:sentry-environment=production,sentry-release=4.1.1,sentry-public_key=xxxxx,sentry-trace_id=xxxxx,sentry-sample_rate=1,sentry-transaction=debug-id-sourcemap-upload,sentry-sampled=true" "--header" "sourcemaps" "upload" "--release" "xxxxx" "/tmp/sentry-bundler-plugin-upload-NEOt49" "--ignore" "node_modules" "--dist" "xxxxx" "--no-rewrite"
```

### Workaround
Currently, the only workaround is to disable source map uploads in the plugin and run `sentry-cli releases files upload-sourcemaps` manually with `--header`.

liAmirali:
Closes #796 
- Problem: The `headers` option is ignored in the `sentryWebpackPlugin`.
- Solution: Pass `options.headers` to `SentryCli` in `createCliInstance` method.
- Impact: Enables use of custom custom headers

### DIFF
--- a/packages/bundler-plugin-core/src/build-plugin-manager.ts
+++ b/packages/bundler-plugin-core/src/build-plugin-manager.ts
@@ -100,6 +100,7 @@ function createCliInstance(options: NormalizedOptions): SentryCli {
     vcsRemote: options.release.vcsRemote,
     headers: {
       ...getTraceData(),
+      ...options.headers,
     },
   });
 }


chargome (approved):
Thanks!

Issue #816, opened 2025-10-22
`@sentry/vite-plugin` (with default `telemetry: true`) used in conjunction with `@module-federation/vite` produces broken builds


joshacheson:
Setting `telemetry: false` remedies the issue, but I imagine others struggle with this problem and would not think to do this.

Minimal reproduction and more context: https://github.com/joshacheson/sentry-vite-module-federation-bug-repro

PR #818, opened 2025-10-24
fix(core): Stop awaiting build start telemetry to avoid breaking module federation builds

andreiborza:
Prior to this fix, we would create a span when the build starts and await flushing it before proceeding with the build. This breaks when using module federation builds as outlined in #816.

This fix removes the blocking await and optimistically fires off the span. This **should** be fine as we definitely care more about not breaking user builds than telemetry arriving. It could lead to us potentially missing this span sometimes when builds run very fast and finish before the span flushes out, however given that the build process usually takes longer (especially when sourcemaps and uploading sourcemaps is involved) it should be fine most of the time.

Closes: #816

### DIFF
--- a/packages/bundler-plugin-core/src/index.ts
+++ b/packages/bundler-plugin-core/src/index.ts
@@ -83,7 +83,7 @@ export function sentryUnpluginFactory({
         // We cannot await the flush here because it would block the build start
         // which in turn would break module federation builds, see
         // https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/816
-        sentryBuildPluginManager.telemetry.emitBundlerPluginExecutionSignal().catch(() => {
+        void sentryBuildPluginManager.telemetry.emitBundlerPluginExecutionSignal().catch(() => {
           // Nothing for the users to do here. If telemetry fails it's acceptable.
         });
       },


packages/bundler-plugin-core/src/index.ts
```diff
@@ -77,8 +77,15 @@ export function sentryUnpluginFactory({
...
       name: "sentry-telemetry-plugin",
-      async buildStart() {
-        await sentryBuildPluginManager.telemetry.emitBundlerPluginExecutionSignal();
+      buildStart() {
+        // Technically, for very fast builds we might miss the telemetry signal
+        // but it's okay because telemetry is not critical for us.
+        // We cannot await the flush here because it would block the build start
+        // which in turn would break module federation builds, see
+        // https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/816
+        sentryBuildPluginManager.telemetry.emitBundlerPluginExecutionSignal().catch(() => {
```
JPeer264: super-l: To highlight that we opted out of awaiting we can add a `void` infront:

```suggestion
        void sentryBuildPluginManager.telemetry.emitBundlerPluginExecutionSignal().catch(() => {
```

PR #823, opened 2025-11-20
chore(deps): Update glob to 10.5.0

### LINKED ISSUE #822: glob vulnerability in @sentry/bundler-plugin-core
See security report for more info: https://github.com/advisories/GHSA-5j98-mcp5-4vw2

`@sentry/bundler-plugin-core` currently resolves to the following:
https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/31d9b6782fb9990c31405e577133640240cefa0f/packages/bundler-plugin-core/package.json#L60

Could we please update to a patched version and release a new version? Thanks.

JPeer264:
closes #822 

We were not affected, since v9 was not part of it. However, I updated this packages just in case. v10.5.0 is the latest version which is supporting v18, which we use here. To not touch too much code I used this version (which is also in the range of the fixed versions)

### DIFF
--- a/packages/bundler-plugin-core/package.json
+++ b/packages/bundler-plugin-core/package.json
@@ -57,7 +57,7 @@
     "@sentry/cli": "^2.57.0",
     "dotenv": "^16.3.1",
     "find-up": "^5.0.0",
-    "glob": "^9.3.2",
+    "glob": "^10.5.0",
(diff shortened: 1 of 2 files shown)


andreiborza (approved):
thanks for taking care of it!

PR #832, opened 2026-01-12
fix(vite): Ensure sentryVitePlugin always returns an array of plugins

nicohrubec:
`sentryVitePlugin` is typed to return `VitePlugin[]`. The problem is that in case the factory we use returns an array with only a single entry, `unplugin` will return a single plugin instead of an array. This can happen when the factory bails early and returns only the `sentry-noop-plugin`.

In other plugins this doesn't seem to be an issue, since we use `any` as the return type there.


Lms24:
argh should have read the pr description ... should have read the pr description ... thanks Nico!

PR #836, opened 2026-01-15
fix(plugin-manager): Respect `sourcemap.ignore` values for injecting debugIDs

s1gr1d:
In [this PR](https://github.com/getsentry/sentry-javascript/pull/18561), Nuxt will use the build plugin manager to upload source maps at the end of the build. Currently, it fails because the `ignore` option is not respected. 

`injectDebugIds` is currently using the "raw" `execute` function and doesn't pass the `--ignore` param yet. The functions in the CLI package contain some defaults around what is passed to `execute`, for example a default `ignore`:
https://github.com/getsentry/sentry-cli/blob/9f34df859bf2975155aad0fe5c6c423b45c07084/lib/releases/index.ts#L16

The next step is to bring this `inject` command to the CLI code ([CLI PR](https://github.com/getsentry/sentry-cli/pull/3088)) so it can be used in the BuildPluginManager. 

Related to https://github.com/getsentry/sentry-javascript/issues/18519

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -1,0 +1,18 @@
+name: Changelog Preview
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+      - labeled
+      - unlabeled
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  changelog-preview:
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    secrets: inherit
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,14 +3,18 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: Version to release
-        required: true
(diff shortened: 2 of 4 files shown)


packages/bundler-plugin-core/src/utils.ts
```diff
@@ -408,3 +408,23 @@ export function getProjects(project: string | string[] | undefined): string[] |
...
   return undefined;
 }
+
+/**
+ * Inlined functionality from @sentry/cli helper code to add `--ignore` options.
+ *
+ * Temporary workaround until we expose a function for injecting debug IDs. Currently, we directly call `execute` with CLI args to inject them.
+ */
+export function serializeIgnoreOptions(ignoreValue: string | string[] | undefined): string[] {
+  const DEFAULT_IGNORE = ["node_modules"];
```
Lms24: hmm I wonder if including node_modules has any implications but I don't _think_ so: We inject debug ids into emitted bundles, meaning the node_modules required by the bundle should already be in there. 

Also, the ignore option for upload also defaults to `node_modules`, right?

PR #844, opened 2026-01-16
feat: Combine injection plugins

timfish:
While benchmarking large projects with Rolldown, I found that `MagicString.generateMap()` has a large overhead. There will be [ways to improve this](https://github.com/rolldown/rolldown/pull/7943#issuecomment-3765006913) in the future, however this issue is exacerbated by the fact that we inject up to 3 separate times via 3 separate plugins. Each of these 3 injects into the top of the source and then has to rewrite the sourcemap!

This PR combines the 3 injection plugins into a single plugin which means the chunks only need the sourcemaps generation once. 

This PR retains esbuild with 3 separate plugins because I haven't worked out how to migrate that yet.

With an [example app of 10k modules](https://github.com/rolldown/benchmarks) and 10MB output file:
|  | Time  |
|---|---|
| Rolldown | 900ms |
| Rolldown + Sentry Rollup plugin | 4.5s |
| Rolldown + Sentry Rollup plugin + these changes | 1.8s |


andreiborza (approved):
Nice!

PR #848, opened 2026-01-21
fix: Inject component annotations into HTML elements rather than React components 

### LINKED ISSUE #492: reactComponentAnnotation.  I use an unintended value for a prop.
### Problem Statement

Hello, I am a software developer using @sentry/vite-plugin.

I am using `reactComponentAnnotation` in `sentryVitePlugin`.

I was thinking of injecting the corresponding `value` into `html's attribute` through `babel`.

However, I saw `data-sentry-component` and `data-sentry-source-file` in `react component prop`,
Because of this value I faced error.

I think this function interferes with the `spread` operation, which is a basic function of javascript.
Because of this, I argue that this feature should be changed.

---------------- Additional explanation ---------------------

I used sentry library in my project.

The packaged version of the library used in my project.

"vite": "^5.4.10",
"@sentry/react": "^8.35.0",
"@sentry/vite-plugin": "^2.22.6",
Sentry's reactComponentAnnotation feature injects properties like data-sentry-component and data-sentry-source-file into component props.

This can cause unintended errors when using methods like Object.keys() to process props, as unexpected keys are included.

For example, in logic where components are dynamically rendered based on specific keys, these additional properties can lead to unintended behavior.

```tsx=
const testObjValue = {
  a: ComponentA,
  b: ComponentB,
};

type ComponentType = {
  a: any;
  b: any;
};

const Component = (props: ComponentType) => {
  const keys = Object.keys(props) as ['a', 'b'];
  return (
    <ul>
      {keys.map((key) => {
        const ChildComponent = testObjValue[key];
        return <ChildComponent key={key} />; 
      })}
    </ul>
  );
};
```

### Solution Brainstorm

The `reactComponentAnnotation` feature of Sentry should be modified to directly add attributes only to the HTML tags in JSX.  
This approach avoids modifying the component's props and affects only the HTML DOM, effectively preventing unintended side effects.

timfish:
- Closes #492

Rather than inject attributes into React components which can cause incompatibilities with props, this PR changes the plugin to instead inject attributes into every HTML element in the component root.

## Downsides

Potentially larger bundle sizes because the attributes will get injected into more elements.

### DIFF
--- a/packages/babel-plugin-component-annotate/src/constants.ts
+++ b/packages/babel-plugin-component-annotate/src/constants.ts
@@ -30,6 +30,29 @@ export const KNOWN_INCOMPATIBLE_PLUGINS = [
   "@react-navigation",
 ];
 
+export const REACT_NATIVE_ELEMENTS: string[] = [
+  "Image",
+  "Text",
+  "View",
+  "ScrollView",
+  "TextInput",
+  "TouchableOpacity",
+  "TouchableHighlight",
+  "TouchableWithoutFeedback",
+  "FlatList",
+  "SectionList",
+  "ActivityIndicator",
+  "Button",
+  "Switch",
+  "Modal",
+  "SafeAreaView",
+  "StatusBar",
+  "KeyboardAvoidingView",
+  "RefreshControl",
+  "Picker",
+  "Slider",
+];
+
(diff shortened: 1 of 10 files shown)


s1gr1d (approved):
LGTM but as discussed offline, we can put this behind a flag to reduce bundle size impact

Issue #861, opened 2026-02-02
Remove unplugin dependency


chargome:
Remove the unplugin dependency and implement native bundler plugins instead.

Unplugin pins us to specific Node versions; removing it also allows us to use newer bundler features without dropping Webpack 4 support.

**Rollup PoC:** [https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/858](<https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/858>)

PR #870, opened 2026-02-10
refactor: webpack no longer uses unplugin

timfish:
- Ref #861

This PR converts the webpack plugin to be a pure webpack plugin without using unplugin.


chargome (approved):
Generally LGTM, just left one question. Feel free to merge if I misunderstood this.

PR #871, opened 2026-02-10
refactor: esbuild no longer uses unplugin

timfish:
- Ref #861

This PR removes unplugin as dependency for the esbuild plugin and instead just exports a regular esbuild plugin.


chargome (approved):
Looks good!

PR #882, opened 2026-02-19
test: Migrate to Vitest

timfish:
This PR:
- Migrates all tests to Vitest
- I had to change the CI cache keys because changing a `package.json` resulted in no lock change and a build failure
- Test usages of the Vite plugin had to add `as unknown as vite.Plugin` because the Vite versions varied between the plugin and integration tests. This does not impact end users because it will use their Vite version.


timfish:
I've fixed the e2e snapshots so this should be ready to merge!

PR #884, opened 2026-02-19
refactor: Remove `glob` usage for Rollup plugin

timfish:


stianjensen:
This would be very nice!

Currently getting the following warning during install:
> warning @sentry/vite-plugin > @sentry/bundler-plugin-core > glob@9.3.5: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me

PR #885, opened 2026-02-19
chore!: updating minimatch

isaacs:
Add minimatch@10.2.2 dev dep in the root to push outdated versions into duplicates, and local packages were updated to use minimatch v10.

The following packages depend on minimatch v3:

- @eslint/eslintrc
- @humanwhocodes/config-array
- @jest/core
- copy-concurrently
- eslint
- eslint-plugin-react
- jest-circus
- jest-config
- jest-runner
- move-concurrently
- rimraf
- terser-webpack-plugin
- test-exclude

Finally, `@rollup/plugin-commonjs` depends on minimatch 5.1.6.

A fix will be backported to resolve the ReDOS on v3 and v5, which can then be updated here.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3536,9 +3536,9 @@ brace-expansion@^2.0.1:
     balanced-match "^1.0.0"
 
 brace-expansion@^5.0.2:
-  version "5.0.2"
-  resolved "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.2.tgz#b6c16d0791087af6c2bc463f52a8142046c06b6f"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -29,10 +29,11 @@
     "@nrwl/cli": "14.5.10",
     "@nrwl/workspace": "14.5.10",
     "lerna": "^6.6.2",
+    "minimatch": "^10.2.2",
+    "npm-run-all": "^4.1.5",
```
chargome: Was this added on purpose?

PR #886, opened 2026-02-20
feat!: Remove support for Node < v18 and webpack v4

timfish:
This PR bumps our minimum support Node version to v18. 

The result is that we have to drop webpack v4 which doesn't support Node >= v18


.github/dependency-review-config.yml
```diff
@@ -0,0 +1,4 @@
+fail-on-severity: "critical"
```
JPeer264: l: I would vote to use `"high"` here and use [the default](https://github.com/getsentry/.github/blob/eb07400c733a23833e6995a52e9dbccc5a9013a2/.github/workflows/dependency-review.yml#L33)

PR #901, opened 2026-03-25
fix(esbuild): Avoid warning when entry point has no default export

### LINKED ISSUE #747: Esbuild: plugin throws a warning when the entry point doesn't have a default export
### Environment

- `node`: 23.9.0
- `esbuild`: 0.25.5
- `@sentry/esbuild-plugin`: 3.5.0

### Steps to Reproduce

Just added the `@sentry/esbuild-plugin`:
```javascript
import 'dotenv/config';
import esbuild from 'esbuild';
import { sentryEsbuildPlugin } from '@sentry/esbuild-plugin';

esbuild.build({
  entryPoints: ['index.ts'],
  sourcemap: true,
  bundle: true,
  minify: true,
  platform: 'node',
  outdir: 'dist',
  allowOverwrite: true,
  plugins: [
    sentryEsbuildPlugin({
      authToken: process.env.SENTRY_AUTH_TOKEN,
      org: process.env.SENTRY_ORG,
      project: process.env.SENTRY_PROJECT,
      sourcemaps: {
        filesToDeleteAfterUpload: ['**/*.js.map'],
      },
    }),
  ],
});
```
If I comment out the plugin the warning is not being displayed. My entry point `index.ts` file does not have any exports. It's an express.js app so I don't have anything to export.

### Expected Result

Not see a warning

### Actual Result

I'm getting the following warning when I'm running the build:
```text
▲ [WARNING] Import "default" will always be undefined because there is no matching export in "index.ts" [import-is-undefined]

    index.ts?sentryProxyModule=true:4:44:
      4 │               export default OriginalModule.default;
        ╵                                             ~~~~~~~
```

queelius:
## Summary

Fixes #747

When using `@sentry/esbuild-plugin` with an entry point that has no default export (common for Express.js apps), the plugin's debug ID proxy module generates:

```js
export default OriginalModule.default;
```

This causes esbuild (0.25+) to emit:

```
▲ [WARNING] Import "default" will always be undefined because there is no matching export in "index.ts" [import-is-undefined]
```

The fix wraps the default export access in a function call, which prevents esbuild's static analysis from detecting the potentially missing `.default` property while still correctly forwarding it when present:

```js
var _sentry_default_export = (function(m) { return m.default; })(_sentry_original_module);
export { _sentry_default_export as default };
```

## Test plan

- [x] Added unit tests covering: no default export (no warning), with default export (preserved), named-only exports (preserved)
- [x] All existing unit tests pass (`vitest run` in `packages/esbuild-plugin`)
- [x] All 86 integration tests pass (`yarn test` in `packages/integration-tests`)
- [x] Lint passes clean (`eslint ./src ./test`)
- [x] Manually verified fix eliminates the warning with esbuild 0.25.5 (version from issue report)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


timfish:
I have concerns about the side-effects this adds to every bundle

PR #908, opened 2026-04-02
feat(core): Pass `mapDir` to `rewriteSourcesHook`

chargome:
- Adds a context parameter (with a mapDir property) to the rewriteSources hook, so users can resolve source paths relative to the source map's location.

ref https://github.com/getsentry/sentry-javascript/issues/20028


timfish (approved):
PR is good but don't sourcemaps contain a relative path to the source files?

Or is this feature for when that is missing or wrong?

chargome:
This is basically just a helper for users that need an anchorpoint for their relative paths like in https://github.com/getsentry/sentry-javascript/issues/20028

Lms24 (approved):

PR #922, opened 2026-05-01
test: Remove old integration tests

timfish:
- This should be merged first for a much smaller diff #921 

This PR removes the older integration tests and fixes a few type issues caused by the removed dependencies.


packages/integration-tests-next/fixtures/esbuild/utils.ts
```diff
@@ -37,7 +37,7 @@ export function test(url: string, callback: TestCallback) {
   // Detect CJS config files by test name suffix
   const configExt = testName.endsWith("-cjs") ? ".config.cjs" : ".config.js";
 
-  vitestTest(`esbuild > ${testName}`, (ctx) =>
+  vitestTest(`pnpm esbuild > ${testName}`, (ctx) =>
```
chargome: this rename was intentional?

chargome (approved):

packages/integration-tests-next/fixtures/esbuild/utils.ts
timfish: Nope, good catch!

PR #923, opened 2026-05-01
feat: Update Sentry CLI to v3

### LINKED ISSUE #825: Update to Sentry CLI v3
Sentry CLI released a new major ([3.0.0](https://github.com/getsentry/sentry-cli/releases/tag/3.0.0)). We should update our plugins to also use this new major.

Since the CLI in v3 limits support for older Self-Hosted versions and sets a ~25.x (TBD)~ 24.11.1 version as the minimum version, we should also cut a major of the plugins to clearly communicate that the new version of the plugins will have the same SH version range support.

(This will further trickle down to SDKs but that's out of scope for this issue/repo).

timfish:
- Closes #825

This PR updates the CLI to v3 which involved:
- Updated CLI exports
- `live` argument has been either removed or changed to `boolean`
- Snapshots now include passing of projects via `-p` flag for new release


timfish:
This issues suggests that this is a breaking change?
- https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/825

Issue #835, opened 2026-01-15
Make plugin return types more specific


nicohrubec:
At the moment the return types for all but the `vite-plugin` is typed as `any`. It would probably be good to use a more specific type here (for instance, the `vite-plugin` uses `VitePlugin[]`.

PR #928, opened 2026-05-11
fix(vite): avoid version-specific plugin return type

### LINKED ISSUE #926: sentryVitePlugin() return type incompatible with Vite 7
## Environment

* `@sentry/vite-plugin`: 5.2.1
* `vite`: 7.3.1
* `typescript`: 5.x

## Description

`sentryVitePlugin()` returns `Plugin[]` (using Vite's `Plugin` type). In Vite 7, the `Plugin` interface changed -- specifically the `hotUpdate` hook's `this` context type (`HotUpdatePluginContext` vs `MinimalPluginContext & { environment: DevEnvironment }`), and `DevEnvironment` gained a required `options` property.

This causes a TypeScript error when using `sentryVitePlugin()` in a `defineConfig` plugins array with Vite 7:

```
error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type 'Plugin<any>[]' is not assignable to type 'PluginOption'.
      ...
      Property 'options' is missing in type 'DevEnvironment' (vite 7)
      but required in type 'DevEnvironment' (vite 6).
```

## Workaround

```ts
import type { PluginOption } from 'vite';

sentryVitePlugin({ ... }) as unknown as PluginOption,
```

## Expected behavior

`sentryVitePlugin()` should return a type compatible with Vite 7's `PluginOption`. This may require bumping the `vite` dependency used for type declarations, or using a version-agnostic return type.

## Related

* [#695](<https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/695>) (closed) -- previously fixed the `any` typing, but the fix used Vite 6's `Plugin` type which is now incompatible with Vite 7
* getsentry/sentry-javascript-bundler-plugins#835 -- tracks making return types more specific

logaretm:
Avoids exposing Vite's version-specific Plugin type from `sentryVitePlugin` by exporting a vite-version-agnostic shape of it.

Also added a Vite 6 fixture and type compatibility coverage for Vite 6, 7, and 8 so we can catch this in the future if it happens again.

Fixes #926

### DIFF
--- a/packages/integration-tests-next/fixtures/vite-type-compat.test.ts
+++ b/packages/integration-tests-next/fixtures/vite-type-compat.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 import * as ts from "typescript";
-import { isAbsolute, join, relative } from "node:path";
+import { isAbsolute, join, normalize, relative } from "node:path";
 import { fileURLToPath } from "node:url";
 import { createRequire } from "node:module";
 
@@ -33,6 +33,7 @@ function assertFixtureViteVersion(fixtureDir: string, expectedMajor: string): vo
 function getDiagnosticsForFixture(fixtureName: string, expectedMajor: string): string[] {
   const fixtureDir = join(fixturesDir, fixtureName);
   const fileName = join(fixtureDir, "sentry-vite-plugin-type-compat.mts");
+  const isVirtualConfigFile = (path: string) => normalize(path) === normalize(fileName);
 
   assertFixtureViteVersion(fixtureDir, expectedMajor);
   assertFixtureViteVersion(pluginViteTypesFixtureDir, "6");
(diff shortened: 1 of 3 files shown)


logaretm:
@nicohrubec Does this also address your concern with #835?

timfish (approved):

timfish:
> Does this also address your concern with https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/835?

The webpack and rollup plugin functions still return `any`:
https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/1176013977d644ac0747dc59423d5a3df666033d/packages/rollup-plugin/src/index.ts#L257
https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/e924c1256108bab65b4abb645ff3a15a60b90f21/packages/webpack-plugin/src/index.ts#L9

I think we decided these couldn't be changed without it being a breaking change.

nicohrubec (approved):

PR #930, opened 2026-05-12
chore: fix craft config

timfish:
The config was wrong because it was copied from a repo using an older craft version


andreiborza (approved):
🙇

PR #931, opened 2026-05-21
chore: Remove `ts-node` usage

timfish:
Just stripping out stuff we've already removed from the JavaScript repository!


packages/dev-utils/rollup.config.mjs
```diff
@@ -0,0 +1,15 @@
+import packageJson from "./package.json" with { type: "json" };
+import modulePackage from "module";
```
chargome:
```suggestion
```

Do we need this?

packages/playground/scripts/request-logger-proxy.mjs
```diff
@@ -29,7 +29,7 @@ app.use(function (req, res, next) {
   var oldWrite = res.write,
     oldEnd = res.end;
 
-  res.write = function (chunk: Buffer) {
+  res.write = function (chunk) {
     resBody.push(chunk);
 
     // @ts-ignore
```
chargome: Some ts-ignores in here that we can probably get rid of

chargome (approved):
Looks good, just minor nits

PR #933, opened 2026-05-21
chore: Align lint configuration with JavaScript repo

timfish:
This PR is built on top of #932 which should be merged first.

This PR:
- Copies the oxlint config from the js repo
- Adds the same version of oxlint dependencies
- Lints the entire repo from the root
- Removes all eslint config and calls
- Fixes or ignores all the additional linting issues oxlint surfaces


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "extends": "@sentry-internal/typescript/tsconfig.json",
+  "include": ["./**/*.ts"],
+  "exclude": ["node_modules"],
+  "compilerOptions": {
+    "module": "ESNext",
+    "moduleResolution": "bundler",
```
chargome, line 8: q: That is intentional right?

chargome (approved):

tsconfig.json
timfish: #932 needs merging before this PR because the TypeScript changes are in there. After that, this PR is just linting changes.

> That is intentional right?

Yes, the TypeScript in this repo gets bundled with Rolldown. Without `bundler` module resolution we get many of:
```
 Module '"X' can only be default-imported using the 'allowSyntheticDefaultImports' flag
```

PR #934, opened 2026-05-22
test: Fix CI caching on Windows

timfish:
There were intermittent integration test failures in CI on Windows. This was because the `*/fixtures/node_modules/*` directories were getting cached and pnpm on Windows uses copies rather than links. 

We don't want any of the fixtures dependencies to be cached anyway!


antonis (approved):
LGTM and the CI is 🟢 🚀

PR #937, opened 2026-06-03
feat: Update Sentry SDK for telemetry 

timfish:
This PR updates the Sentry SDK used to send telemetry to v10.

The SDK is already bundled into the bundler plugins core module.


isaacs (approved):
👍

PR #938, opened 2026-06-04
feat: Add `@sentry/bundler-plugins` core package

timfish:
Consolidates all bundler plugin code into a single new `@sentry/bundler-plugins` package with subpath exports (`/rollup`, `/webpack`, `/vite`, etc.). Code was moved using `git mv` to preserve history.

The existing bundler-specific packages (`@sentry/rollup-plugin`, etc.) become thin stubs that re-export from `@sentry/bundler-plugins`. 

This makes it straightforward to:

- Add new bundler support (e.g. rolldown) without publishing a new package
- Move all plugin code in the JS monorepo as a single dependency rather than 6
- Keep bundler-specific packages at v5, avoiding unnecessary major version bumps for users

Individual packages can eventually be deprecated but there's no urgency — integration tests still target them and they remain fully functional.


isaacs (approved):
:+1:

PR #940, opened 2026-06-05
fix(webpack): Make `webpack` import lazy to support rspack-only projects

### LINKED ISSUE #939: Support rspack only projects
Hello, on rspack only projects when we use the plugin `@sentry/webpack-plugin` we get the following error during build:

```
Error: Cannot find module 'webpack'
  at /node_modules/@sentry/webpack-plugin/dist/cjs/index.js
```

## Suggestion
 `// src/index.ts`

Fall back to `@rspack/core` if webpack is not found or something similar. 

```ts

let webpack4or5: any = {};

try {
  webpack4or5 = require("webpack");
} catch {
  try {
    webpack4or5 = require("@rspack/core");
  } catch {}
}

const BannerPlugin = webpack4or5?.BannerPlugin || webpack4or5?.default?.BannerPlugin;
const DefinePlugin = webpack4or5?.DefinePlugin || webpack4or5?.default?.DefinePlugin;
```

And mark both as optional peer dependencies

```json
"peerDependenciesMeta": {
  "webpack": { "optional": true },
  "@rspack/core": { "optional": true }
}
```

Would that be possible?

s1gr1d:
The `./webpack` entry has a top-level `import * as webpack from "webpack"` that compiles to an unconditional `require("webpack")`. In rspack-only projects where the `webpack` package isn't installed, this throws `Cannot find module 'webpack'` at module load (before the plugin even runs).

This PR replaces the static import with a guarded `createRequire` call that catches the error and falls back to `undefined`. This is safe because for webpack 5.1+ and rspack, `BannerPlugin`/`DefinePlugin` are already sourced from `compiler.webpack` at runtime ([see here](https://github.com/webpack/webpack/commit/65eca2e529ce1d79b79200d4bdb1ce1b81141459#diff-732395f8fcd972919381425c80194bb0def0a2af50c0940762299605a81c0006R174)). The top-level import only served as a fallback for webpack < 5.1.

The `./webpack5` entry was already unaffected since it calls `sentryWebpackPluginFactory()` with no arguments and relies entirely on `compiler.webpack`.

closes https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/939

### DIFF
--- a/packages/bundler-plugins/src/webpack/index.ts
+++ b/packages/bundler-plugins/src/webpack/index.ts
@@ -1,10 +1,32 @@
 import type { SentryWebpackPluginOptions } from "./webpack4and5";
 import { sentryWebpackPluginFactory } from "./webpack4and5";
-import * as webpack4or5 from "webpack";
+import { createRequire } from "node:module";
 
-const BannerPlugin = webpack4or5?.BannerPlugin || webpack4or5?.default?.BannerPlugin;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type PluginClass = new (options: any) => unknown;
 
-const DefinePlugin = webpack4or5?.DefinePlugin || webpack4or5?.default?.DefinePlugin;
+type WebpackModule = {
+  BannerPlugin?: PluginClass;
+  DefinePlugin?: PluginClass;
+  default?: WebpackModule;
+};
+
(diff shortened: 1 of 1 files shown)


s1gr1d:
@bugbot review

timfish (approved):

PR #944, opened 2026-06-27
perf(vite): add fast component annotation transform

### LINKED ISSUE #943: Vite 8 /Rolldown: Do not use Babel for annotating react components
### Steps to Reproduce

1. Create a Vite + React project and add a good amount of JSX components
2. Build the project. This is fast
3. Enable `reactComponentAnnotation` 
4. Build again - now it is slow :( 

### Actual Result

When using the option in Sentry's Vite plugin, a Babel transform is utilized which slows down builds with a lot of components.

https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/988efd30691e08c059eb577e499d0b4346434f3c/packages/bundler-plugins/src/rollup/index.ts#L113

### Expected Result

A Rolldown/Vite 8 parse + Magic String transform is used which should take a fraction of the time.

TheAlexLichter:
Resolves #943

This PR adds a port of the Babel transform for`reactComponentAnnotation.enabled` that works with Vite 8 + magic string, dramatically decreasing build time.

I've also added parity tests and consolidated values to avoid drift of both implementations.

Vite 7 and fallback cases will use the existing Babel transform.
Both Babel and the Vite annotation backend are lazy-loaded to avoid unnecessary startup cost.

### DIFF
--- a/packages/bundler-plugins/test/core/component-annotation-vite.test.ts
+++ b/packages/bundler-plugins/test/core/component-annotation-vite.test.ts
@@ -1,0 +1,327 @@
+import { transformAsync, traverse, types as t } from "@babel/core";
+import { parse } from "@babel/parser";
+import MagicString from "magic-string";
+import { describe, expect, it, vi } from "vitest";
+
+import componentNameAnnotatePlugin from "../../src/babel-plugin";
+import {
+  createViteComponentNameAnnotateHooks,
+  type ComponentAnnotationTransformResult,
+} from "../../src/core/component-annotation-vite";
+
+type Annotation = {
+  elementName: string;
+  attributes: Record<string, string>;
+};
+
+const SENTRY_ATTRIBUTES = new Set([
+  "data-sentry-component",
+  "data-sentry-element",
+  "data-sentry-source-file",
+]);
+
+async function parseAstAsync(code: string, options: { lang: "jsx" | "tsx" }): Promise<unknown> {
+  return parse(code, {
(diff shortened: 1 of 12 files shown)


packages/bundler-plugins/test/core/component-annotation-vite.test.ts
```diff
@@ -0,0 +1,327 @@
+import { transformAsync, traverse, types as t } from "@babel/core";
+import { parse } from "@babel/parser";
```
TheAlexLichter, line 2: Yes, this does use babel for testing as it is more convenient API-wise

packages/integration-tests-next/fixtures/vite8/component-annotation.test.ts
```diff
@@ -39,13 +39,13 @@ test(import.meta.url, ({ runBundler, readOutputFiles }) => {
     			"data-sentry-source-file": "app.jsx"
     		}, void 0, false, {
     			fileName: _jsxFileName,
-    			lineNumber: 4,
+    			lineNumber: 6,
```
TheAlexLichter, line 42: The updated line numbers are on purpose and actually "more correct" as they now point to the actual JSX component, not just the span around it.

timfish:
Closing in favour of [this PR](https://github.com/getsentry/sentry-javascript/pull/22089) in the JavaScript repository.

Apologies for leaving this ambiguity lying around in this repo!

PR #946, opened 2026-07-08
test: Remove versions from telemetry snapshots

timfish:
We shouldn't be asserting the SDK and Node versions in the tests!


Lms24 (approved):
thanks!

